PR #2, opened 2020-03-19
Expose "ready" property for checking if server installed successfully

rchl:
The idea is to then have a check like this one in `def config`  in LSP-* packages:
```py
        if not server.ready:
            return read_client_config(self.name, {
                'enabled': False,
                'languages': [{}],
            })
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.mypy_cache
--- a/st3/lsp_utils/server_npm_resource.py
+++ b/st3/lsp_utils/server_npm_resource.py
@@ -43,10 +43,15 @@ class ServerNpmResource(object):
 
     def __init__(self, package_name, server_directory, server_binary_path):
         self._initialized = False
+        self._is_ready = False
         self._package_name = package_name
         self._server_directory = server_directory
         self._binary_path = server_binary_path
         self._package_cache_path = None
+
+    @property
+    def ready(self):
+        return self._is_ready
 
     @property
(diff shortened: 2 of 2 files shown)


rchl:
Actually this won't work that well as once LSP gets such configuration, it will consider client disabled for session so restart would be necessary.

Would probably need some bigger refactoring to get this working as expected.

@rwols FYI

rwols:
The way to solve this for now IMO is to just run the code in the main thread, not in a separate thread.

With asyncio we could make this non-blocking, but that's a story for later.

rchl:
No, sorry, I'd rather live with current (minor) problem as blocking the main thread would give a horrible user experience (especially with bigger servers like lsp-vue).

rwols:
Sure :)

rchl:
Actually still gonna go with this change as we can instead bailout in `on_start` override, based on `server.ready`.

predragnikolic:
Your intent is to use the `.ready` boolean in the `on_start` method of a LanguageHandler
to not start the LanguageHandler if the server is not ready. Is that right?

rchl:
> Your intent is to use the `.ready` boolean in the `on_start` method of a LanguageHandler
> to not start the LanguageHandler if the server is not ready. Is that right?

Correct

predragnikolic:
Can we make a release now? :)

rchl:
Sure :)

rchl:
1.1.0 released

PR #4, opened 2020-03-29
Show nicer progress indicator when installing server

rchl:


predragnikolic:
But what happens when you are installing two servers at the same time?

rchl:
> But what happens when you are installing two servers at the same time?

`label` is the key so you get both displayed.

predragnikolic:
The label is not a key, it will be shown just next to the animation :)
>     If `label` is provided, then it will be shown next to the animation.

By looking at the sublime_lib source
If we pass in a view instead of the `sublime.active_window`
the key(UUID) will be automaticly generated.
```
class ActivityIndicator:
    ...
    def __init__(
        self,
        target: Union[StatusTarget, sublime.View, sublime.Window],
        label: Optional[str] = None,
    ) -> None:
        self.label = label


        if isinstance(target, sublime.View):
            self._target = ViewTarget(target)
        elif isinstance(target, sublime.Window):
            self._target = WindowTarget(target)
        else:
            self._target = target

...
    class ViewTarget(StatusTarget):
    def __init__(self, view: sublime.View, key: Optional[str] = None) -> None:
        self.view = view
        if key is None:
            self.key = '_{!s}'.format(uuid4())
        else:
            self.key = key
```

but because we are passing it a Window class not a VIew.
this function here will be called
https://github.com/SublimeText/sublime_lib/blob/e8aeb59eacb14c52836c4210c0b011aaaf81da96/st3/sublime_lib/activity_indicator.py#L30-L31

output

rchl:
Hmm, right, window status doesn't take the key...

Worth handling this edge case somehow? I doubt myself.

We could use some view (emphasis on "some") but I don't think that would be better as the installation is not really related to any specific file - it triggers from global scope.

predragnikolic:
To sum up.

Status message overlaps if multiple LSP-* plugins are installing at the same time.
But on the other side the installation is quick for most packages. 
And the most important thing is that it gives a better feedback to the user that something is happening.

But I will let @rwols decide. :)

rchl:
I've switched to using an active view instead.
That is what [PC also uses](https://github.com/wbond/package_control/blob/ea2c658bf48ded502c2619af3aa70e86dcd2cd00/package_control/thread_progress.py#L44).

PR #5, opened 2020-04-29
Create NpmClientHandler to simplify NPM handler creation

rchl:
Adds `NpmClientHandler` export that at minimium requires only
overriding PACKAGE_NAME, SERVER_DIRECTORY and SERVER_BINARY_PATH class
attributes.

### DIFF
--- a/st3/lsp_utils/npm_client_handler.py
+++ b/st3/lsp_utils/npm_client_handler.py
@@ -1,0 +1,126 @@
+import shutil
+import sublime
+
+from LSP.plugin.core.handlers import LanguageHandler
+from LSP.plugin.core.settings import ClientConfig, read_client_config
+from LSP.plugin.core.typing import Dict
+from .server_npm_resource import ServerNpmResource
+
+CLIENT_SETTING_KEYS = ['env', 'experimental_capabilities', 'languages', 'initializationOptions', 'settings']
+
+
+class NpmClientHandler(LanguageHandler):
+    # To be overridden by subclass.
+    PACKAGE_NAME = None
+    SERVER_DIRECTORY = None
+    SERVER_BINARY_PATH = None
+    # Internal
+    __SERVER = None
+
+    def __init__(self):
+        super().__init__()
+        assert self.PACKAGE_NAME
+        self.package_name = self.PACKAGE_NAME
+        self.settings_filename = '{}.sublime-settings'.format(self.package_name)
+
+        # Calling setup() also here as this might run before `plugin_loaded`.
+        # Will be a no-op if already ran.
+        # See https://github.com/sublimelsp/LSP/issues/899
+        self.setup()
+
+    @classmethod
+    def setup(cls) -> None:
+        assert cls.PACKAGE_NAME
+        assert cls.SERVER_DIRECTORY
+        assert cls.SERVER_BINARY_PATH
+        if not cls.__SERVER:
+            cls.__SERVER = ServerNpmResource(cls.PACKAGE_NAME, cls.SERVER_DIRECTORY, cls.SERVER_BINARY_PATH)
+        cls.__SERVER.setup()
+
+    @classmethod
+    def cleanup(cls) -> None:
+        if cls.__SERVER:
+            cls.__SERVER.cleanup()
+
+    @property
+    def name(self) -> str:
+        return self.package_name.lower()
+
+    @property
+    def config(self) -> ClientConfig:
+        assert self.__SERVER
+
+        configuration = {
+            'enabled': True,
+            'command': ['node', self.__SERVER.binary_path] + self.get_binary_arguments(),
+        }
+
+        configuration.update(self._read_configuration())
+        self.on_client_configuration_ready(configuration)
+        return read_client_config(self.name, configuration)
+
+    def get_binary_arguments(self):
+        """
+        Returns a list of extra arguments to append when starting server.
+        """
+        return ['--stdio']
+
+    def _read_configuration(self) -> Dict:
+        settings = {}  # type: Dict
+        loaded_settings = sublime.load_settings(self.settings_filename)  # type: Dict
+
+        if loaded_settings:
+            migrated = self._migrate_obsolete_settings(settings)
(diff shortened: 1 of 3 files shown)


rchl:
As an example, `LSP-eslint` would like like that when using this class:
```py
import os
import webbrowser

from LSP.plugin.core.protocol import Response
from lsp_utils import NpmClientHandler


def plugin_loaded():
    LspEslintPlugin.setup()


def plugin_unloaded():
    LspEslintPlugin.cleanup()


class LspEslintPlugin(NpmClientHandler):
    PACKAGE_NAME = __package__
    SERVER_DIRECTORY = 'vscode-eslint'
    SERVER_BINARY_PATH = os.path.join(SERVER_DIRECTORY, 'out', 'eslintServer.js')

    def on_initialized(self, client) -> None:
        client.on_notification('eslint/status', self.handle_status)
        client.on_request(
            'eslint/openDoc',
            lambda params, request_id: self.handle_open_doc(client, params, request_id))

    def handle_status(self, params) -> None:
        pass

    def handle_open_doc(self, client, params, request_id) -> None:
        webbrowser.open(params['url'])
        client.send_response(Response(request_id, {}))
```

@jfcherng @deathaxe @RandomByte

rchl:
I've checked all npm-based LSP-* packages and this design should address all of them.

We could abstract it even more - hide the `client` and provide abstractions for `client.on_notification/on_request/send_response` methods. Then, when the LSP API changes, we could address all of that within `lsp_utils` without changing packages (in theory).

rchl:
There is a problem. LSP initializes subclasses of `LanguageHandler` but we don't want to initialize `NpmClientHandler ` (which is a direct subclass of `LanguageHandler`) but only subclasses of `NpmClientHandler`. So this either needs a solution in LSP itself or something else that I can't think of.

deathaxe:
Just out of couriosity: Why do we need a separate dependency to provide a class which is derived from an abstract interface class the LSP package?

I understand its purpose and what it does, but why can't such infrastructure not be part of the LSP package? 

The dependency needs to import from LSP, any LSP-blabla package needs to import the dependency and other classes from LSP. All in all everything requires LSP package to be available first for all that stuff to work.

It feels pretty complicated.

rchl:
> I understand its purpose and what it does, but why can't such infrastructure not be part of the LSP package?

I think that's because as much LSP doesn't want to be concerned with package-specific behaviors and settings, it also wouldn't want to be concerned with various ways of handling their installation and dependency management.

st3/lsp_utils/npm_client_handler.py
```diff
@@ -0,0 +1,146 @@
...
+            lambda params, request_id: self.__client.send_response(Response(request_id, handler(params))))
+
+
+class NpmClientHandler(LanguageHandler):
+    # To be overridden by subclass.
+    PACKAGE_NAME = None
+    SERVER_DIRECTORY = None
+    SERVER_BINARY_PATH = None
+    # Internal
+    __SERVER = None
```
deathaxe, line 32: Why capital letters? It is an object/instance, isn't it?

rchl: yes, was just thinking about it

rchl:
> There is a problem. LSP initializes subclasses of `LanguageHandler` but we don't want to initialize `NpmClientHandler ` (which is a direct subclass of `LanguageHandler`) but only subclasses of `NpmClientHandler`. So this either needs a solution in LSP itself or something else that I can't think of.

Addressing in https://github.com/sublimelsp/LSP/pull/1000

st3/lsp_utils/npm_client_handler.py
```diff
@@ -0,0 +1,148 @@
...
+    package_name = None
+    server_directory = None
+    server_binary_path = None
+    # Internal
+    __server = None
+
+    def __init__(self):
+        super().__init__()
+        assert self.package_name
+        self.package_name = self.package_name
```
deathaxe, line 37: What do you try to achieve by that?

rchl: This is not how the latest code looks like :)

st3/lsp_utils/npm_client_handler.py
```diff
@@ -0,0 +1,157 @@
...
+        client = settings.get('client')  # type: Dict
+        if client:
+            settings.erase('client')
+            # Migrate old keys
+            for key, value in client.items():
+                settings.set(key, value)
+            return True
+        return False
+
+    def on_client_configuration_ready(self, configuration: Dict):
```
predragnikolic, line 131: Sorry for the late reply.
Small nitpick on `on_client_configuration_ready(...)`.
Should we add type annotation for the return value?

`on_client_configuration_ready(self, configuration: Dict) -> None: `
This type annotation means that we don't need to return the configuration.

`on_client_configuration_ready(self, configuration: Dict) -> Dict: `
This type annotation means that we do need to return the configuration.

Just to be explicit :)

st3/lsp_utils/npm_client_handler.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        Called when package settings were read. Receives a `sublime.Settings` object.
+
+        Can be used to change user settings, migrating them to new schema, for example.
+
+        Return True if settings were modified to save changes to file.
+        """
+        return False
+
+    def _migrate_obsolete_settings(self, settings: Dict):
```
predragnikolic, line 116: we can probably deprecate this in the near future

predragnikolic: as I believe that the users who had the `client` settings, had already migrated to the new way.

rchl: Yes, we can. But since it's out of the way now, it shouldn't bother anyone. :)

st3/lsp_utils/npm_client_handler.py
rchl: Fixed in 6fc936b7991a72b66f78a26bf7bde88e80288bcd

PR #8, opened 2020-05-10
Attempt #1 at making both old API and new API work

rwols:
Attempt at making both old API and new API work. Relevant pull request: https://github.com/sublimelsp/LSP/pull/979

This won't account for LSP-lemminx, LSP-eslint, LSP-html (why does this one not use this helper package? Because it's JS?), metals-sublime.

It's a bit weird that the base class differs. Perhaps we can make the LanguageHandler act as the "ApiWrapper". It will be a class that is actually an ApiWrapper and will carry the name "LanguageHandler" only for the sake of the name :)

### DIFF
--- a/st3/lsp_utils/npm_client_handler.py
+++ b/st3/lsp_utils/npm_client_handler.py
@@ -1,11 +1,30 @@
+from .server_npm_resource import ServerNpmResource
+from LSP.plugin.core.typing import Callable, Dict, List, Optional
 import shutil
 import sublime
 
-from LSP.plugin.core.handlers import LanguageHandler
-from LSP.plugin.core.protocol import Response
-from LSP.plugin.core.settings import ClientConfig, read_client_config
-from LSP.plugin.core.typing import Callable, Dict
-from .server_npm_resource import ServerNpmResource
+USE_NEW_API = True
+
+try:
+
+    from LSP.plugin import ClientConfig
+    from LSP.plugin import read_client_config
+    from LSP.plugin import Response
+    from LSP.plugin import Session
+    from LSP.plugin import WorkspaceFolder
+    from LSP.plugin.core.rpc import method2attr  # temporary
+    BaseClass = Session
+
+except ImportError:
+
+    USE_NEW_API = False
+    from LSP.plugin.core.handlers import LanguageHandler
+    from LSP.plugin.core.protocol import Response
+    from LSP.plugin.core.protocol import WorkspaceFolder
+    from LSP.plugin.core.settings import ClientConfig
+    from LSP.plugin.core.settings import read_client_config
+    BaseClass = LanguageHandler
+
 
 # Keys to read and their fallbacks.
 CLIENT_SETTING_KEYS = {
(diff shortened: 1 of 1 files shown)


st3/lsp_utils/npm_client_handler.py
```diff
@@ -1,11 +1,30 @@
...
-from LSP.plugin.core.handlers import LanguageHandler
-from LSP.plugin.core.protocol import Response
-from LSP.plugin.core.settings import ClientConfig, read_client_config
-from LSP.plugin.core.typing import Callable, Dict
-from .server_npm_resource import ServerNpmResource
+USE_NEW_API = True
+
+try:
+
+    from LSP.plugin import ClientConfig
```
rwols, line 10: Let's all shake hands and say "The public API for server handlers is in plugin/\_\_init\_\_.py". Then try to keep that list stable. Importing from anywhere else is not guaranteed to be stable.

rchl:
> This won't account for LSP-lemminx, LSP-eslint, LSP-html (why does this one not use this helper package? Because it's JS?), metals-sublime.

LSP-eslint has some breaking changes that are blocked by code actions on save feature so I can't release it yet and didn't bother updating it.

LSP-html seems updated already.

rchl:
I have a slightly different solution. I'll put a PR for mine and we'll see which one to pick.

st3/lsp_utils/npm_client_handler.py
```diff
@@ -59,51 +92,77 @@ def setup(cls) -> None:
...
+                sublime.save_settings(cls.settings_filename)
 
             for key, default in CLIENT_SETTING_KEYS.items():
                 settings[key] = loaded_settings.get(key, default)
 
         return settings
 
-    def on_settings_read(self, settings: sublime.Settings):
+    @classmethod
+    def on_settings_read(cls, settings: sublime.Settings):
```
rchl, line 165: This is not backwards compatible, is it? Previously it was a method of the instance, now it's a method of the class. The overrides won't work.

rwols: You can call classmethods on an instance just fine :)

```
class Foo:

    @classmethod
    def f(cls):
        print("hello")


foo = Foo()
foo.f()
```

rwols: Also @classmethods can be overridden. @staticmethod can not be overridden. @classmethod is magic.

st3/lsp_utils/npm_client_handler.py
```diff
@@ -17,39 +36,53 @@
...
-        def on_response(params, request_id):
-            handler(params, lambda result: send_response(request_id, result))
+        if USE_NEW_API:
+
+            def handler_wrapper(this, params, request_id):
+                handler(params, lambda result: this.send_response(Response(request_id, result)))
 
-        def send_response(request_id, result):
-            self.__client.send_response(Response(request_id, result))
+            setattr(self.__client, method2attr(method), handler_wrapper)
```
rchl, line 56: Definitively would be good to have a method for registering notification/request handlers.

st3/lsp_utils/npm_client_handler.py
rchl: But this (which is how it would have to be done today) doesn't work:
```py
class A(object):
    @classmethod
    def configuration(cls):
        cls.on_read_configuration()

    @classmethod
    def on_read_configuration(cls):
        return 'Called on_read_configuration'


class B(A):
    def on_read_configuration(self):
        print('Ran overriden on_read_configuration')


b = B()
b.configuration()
```

rwols: The only method that is in use like that is `on_client_configuration_ready`. It is used by LSP-json, LSP-vue and LSP-intelephense.

LSP-vue and LSP-intelephense don't reference instance properties at all, so it's okay to use a hack like this:

```
        called = False
        for c in cls.__subclasses__():
            f = getattr(c, 'on_client_configuration_ready', None)
            if callable(f):
                f(cls, configuration)
                called = True
                break
        if not called:
            cls.on_client_configuration_ready(configuration)
```

For LSP-json we can't do that as it accesses the instance property `self._default_schemas`.

IMO it's okay to rewrite that a bit to make it a classmethod:

```python
class LspJSONPlugin(NpmClientHandler):
    package_name = __package__
    server_directory = 'vscode-json-languageserver'
    server_binary_path = os.path.join(
        server_directory, 'node_modules', 'vscode-json-languageserver', 'bin', 'vscode-json-languageserver'
    )
    _default_schemas = []  # type: List[Dict]

    @classmethod
    def on_client_configuration_ready(cls, configuration: Dict):
        if not cls._default_schemas:
            schemas = ['schemas_extra.json', 'schemas.json']
            for schema in schemas:
                path = 'Packages/{}/{}'.format(cls.package_name, schema)
                cls._default_schemas.extend(json.loads(ResourcePath(path).read_text()))

        configuration['settings'].setdefault('json', {})['schemas'] = cls._default_schemas
```

rchl: I guess this should work https://github.com/sublimelsp/lsp_utils/pull/9

rwols:
I'll re-introduce LanguageHandler as a temporary ApiWrapper class, that might work even for handlers like LSP-lemminx and such.

st3/lsp_utils/npm_client_handler.py
```diff
@@ -59,51 +92,77 @@ def setup(cls) -> None:
...
-            'command': ['node', self.__server.binary_path] + self.get_binary_arguments(),
+            'command': ['node', cls.__server.binary_path] + cls.get_binary_arguments(),
         }
 
-        configuration.update(self._read_configuration())
-        self.on_client_configuration_ready(configuration)
-        return read_client_config(self.name, configuration)
+        configuration.update(cls._read_configuration())
+        cls.on_client_configuration_ready(configuration)
+        return read_client_config(cls.name, configuration)
```
rchl, line 114: That could be simplified to just return configuration maybe?
Calling code can create config itself (it knows the name).

PR #12, opened 2020-05-10
Create separate handler for new API

rchl:
Alternative version to #8
The main difference, I guess, is that it creates it in a separate files rather than merging old and new together.

### DIFF
--- a/st3/lsp_utils/npm_client_handler_v2.py
+++ b/st3/lsp_utils/npm_client_handler_v2.py
@@ -1,0 +1,176 @@
+import shutil
+import sublime
+from LSP.plugin.core.sessions import Session
+from LSP.plugin.core.protocol import Response, WorkspaceFolder
+from LSP.plugin.core.settings import ClientConfig, read_client_config
+from LSP.plugin.core.typing import Callable, Dict, List, Optional
+from .server_npm_resource import ServerNpmResource
+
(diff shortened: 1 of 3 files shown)


st3/lsp_utils/npm_client_handler_v2.py
```diff
@@ -0,0 +1,176 @@
...
+            if migrated or changed:
+                sublime.save_settings(cls.__settings_filename)
+
+            for key, default in CLIENT_SETTING_KEYS.items():
+                settings[key] = loaded_settings.get(key, default)
+
+        return settings
+
+    @classmethod
+    def _migrate_obsolete_settings(cls, settings: sublime.Settings):
```
predragnikolic, line 112: What is the difference between `_migrate_obsolete_settings`and  `on_settings_read`? 

Both can be used to migrate them to new schema, 
but `on_settings_read` can be used to change other settings while `_migrate_obsolete_settings` (probably) should not?

Can you clarify it for me? :)

rchl: `_migrate_obsolete_settings` is an internal function that always runs and can't be overridden. And also can be removed soon because we can assume that all configurations were already migrated.

`on_settings_read` is the one that subclasses can override and do other kind of migration.

PR #13, opened 2020-05-18
Apply AbstractPlugin base class

rwols:
I did not take care of versioning, but this works with eslint.


st3/lsp_utils/npm_client_handler.py
```diff
@@ -152,11 +128,13 @@ def on_start(cls, window) -> bool:
             return False
         return cls.__server.ready
 
-    def on_initialized(self, client) -> None:
+    def __init__(self, weaksession: 'weakref.ref[Session]') -> None:
```
rwols, line 131: Another nice thing about this `AbstractPlugin` is that `__init__` is called at the time when the `Session` is initialized, and `__del__` is called when `shutdown` is about to be sent. So, derivatives from `AbstractPlugin` don't have to care about this initialize/initialized/shutdown/exit dance.

st3/lsp_utils/npm_client_handler.py
```diff
@@ -68,22 +67,14 @@ def cleanup(cls) -> None:
...
-    def config(self) -> ClientConfig:
-        assert self.__server
-
-        configuration = {
-            'enabled': True,
-            'command': ['node', self.__server.binary_path] + self.get_binary_arguments(),
-        }
+    @classmethod
+    def name(cls) -> str:
+        cls.setup()
```
predragnikolic, line 72: The call for setup doesn't feel right here. 

Is there a reason why you called setup in a method that should return a name 🙂

rwols: Because: https://github.com/sublimelsp/LSP/issues/899

st3/lsp_utils/npm_client_handler.py
```diff
@@ -21,39 +26,33 @@ def is_node_installed():
...
-            self.__client.send_response(Response(request_id, result))
+            session = self.__plugin.weaksession()
+            if session:
+                session.send_response(Response(request_id, result))
+
+        def on_response(this, params, request_id):
+            handler(params, lambda result: send_response(request_id, result))
 
-        self.__client.on_request(method, on_response)
+        setattr(self.__plugin, method2attr(method), on_response)
```
rchl, line 45: Can't we hide this behind an API? Why do we need to manually fiddle with `method2attr` function? Could just create `AbstractPlugin.on_request` and `AbstractPlugin.on_notification` that would do that behind the scenes.

st3/lsp_utils/npm_client_handler.py
predragnikolic: Why don't we define a `setup` classmethod on the Abstract class?
```
class AbstractPlugin(metaclass=ABCMeta):
    @classmethod
    @abstractmethod
    def setup(cls) -> None:
        """
        Do the setup
        """
        # not mandatory to implement
        pass 
```
than in register_plugin before returning a Client config call plugin.setup()
```diff

def register_plugin(plugin: Type[AbstractPlugin], update_global_configs: bool = True) -> None:
    global _plugins
    global client_configs

+   plugin.setup()
    try:
        config = ClientConfig(
            name=plugin.name(),
```

Would that work?

rchl: Since `setup` can be async and run on separate thread, it should ideally not be called during `register_plugin`. `register_plugin` should be called only after `setup` has been fully done.

st3/lsp_utils/npm_client_handler.py
rwols: The NpmClientHandler is the API that hides it?

rchl: I'd like this hidden from public API. I imagine some packages will not use `NpmClientHandler`.

st3/lsp_utils/__init__.py
```diff
@@ -1,2 +1,12 @@
-from .npm_client_handler import NpmClientHandler
-from .server_npm_resource import ServerNpmResource
+try:
+    from .v1 import ApiWrapper
+    from .v1 import NpmClientHandler
+except ImportError:
+    from .v0 import ApiWrapper  # type: ignore
+    from .v0 import NpmClientHandler
```
rwols, line 6: Works with both old and new. Some plugins like vscode-html and vue need some more adjustments.

rchl, line 6: I think I'd prefer an explicit variant that does selection based on the version of LSP.
I can imagine in the future there might be some API changes that don't cause the import to fail and that approach would not work then.

rchl: On the other hand I don't imagine us doing such drastic API changes that we would need to create a separate file for them...

st3/lsp_utils/v1.py
```diff
@@ -0,0 +1,120 @@
...
+        return cls.package_name[len('LSP-'):]
+
+    @classmethod
+    def configuration(cls) -> sublime.Settings:
+        if not cls._package_cache_path:
+            # https://github.com/sublimelsp/LSP/issues/899
+            cls.setup()
+        settings = super().configuration()
+        path = os.path.join(sublime.cache_path(), cls.package_name, cls.server_binary_path)
+        settings.set('command', ['node', path, '--stdio'])
```
rchl, line 76: The `--stdio` part can come from `cls.get_binary_arguments()`. `LSP-bash` needs different value.

st3/lsp_utils/__init__.py
rchl: What kind of adjustments? `LSP_html` is as bare as it gets so I'm surprised that it needs something extra.

st3/lsp_utils/v1.py
```diff
@@ -0,0 +1,120 @@
...
+        return True
+
+    @classmethod
+    def install_or_update(cls) -> None:
+        cache_server_path = os.path.join(cls._package_cache_path, cls.server_directory)
+        src_path = 'Packages/{}/{}/'.format(cls.package_name, cls.server_directory)
+        ResourcePath(src_path).copytree(cache_server_path, exist_ok=True)
+        cmd = ["npm", "install", "--verbose", "--production", "--prefix",
+               cls._cache_server_path, cls._cache_server_path]
+        subprocess.check_call(cmd, shell=sublime.platform() == 'windows')
```
rchl, line 102: Would be good to port over some parts from ServerNpmResource:
 - activity indicator
 - some status messages (including on error)

rwols:
Honestly why are we even bothering with all these tiny packages? We could have one large LSP-node-based-language-server-pack that downloads bash/intelephense/css/eslint/html/what-have-you.

st3/lsp_utils/v1.py
rwols: That is something for the base LSP package. Exceptions are already caught in the familiar way by showing an error dialog, same as when your server subprocess can't start.

st3/lsp_utils/v1.py
rwols: Fixed.

st3/lsp_utils/__init__.py
rwols: I got confused because I had not done `git pull` in a while for LSP-html.

st3/lsp_utils/v1.py
```diff
@@ -0,0 +1,143 @@
...
+    def name(cls) -> str:
+        assert cls.package_name
+        return cls.package_name[len('LSP-'):]
+
+    @classmethod
+    def configuration(cls) -> sublime.Settings:
+        if not cls._package_cache_path:
+            # https://github.com/sublimelsp/LSP/issues/899
+            cls.setup()
+        settings = super().configuration()
```
rchl, line 81: It might be a bit annoying to have a `sublime.Settings` object and not a Dict here.

I suppose this is to allow mutating settings but in that case, how would the changes be saved? Would this package have to "guess" settings file name? I mean, it's not hard to guess but it's not explicitly defined here either.

Maybe there should be a separate method (like before) that would get the initial `sublime.Settings` and could mutate it and then the final configuration would be received here as a Dict?

The way it is now also forced you to do the ugly hack with calling `on_client_configuration_ready` twice.

rwols: I'm convinced that getting a `sublime.Settings` object from a plugin is good, because the base LSP package can then listen for changes (add_on_change, clear_on_change). When a change occurs, we can hot-reload in LSP. At least, that's the goal.

> The way it is now also forced you to do the ugly hack with calling on_client_configuration_ready twice.

Well, eventually `on_client_configuration_ready` will disappear :)

> how would the changes be saved? 

Hmm, I don't understand why that would be needed? Why do we need to save changes to the sublime.Settings object to disk?

It's only for listening for changes from the user. If the user saves their overrides in Packages/User/LSP-foo.sublime-settings, the on_change callback is called.

rchl: > Hmm, I don't understand why that would be needed? Why do we need to save changes to the sublime.Settings object to disk?

We've done that before to migrate to new settings structure (the code is still there in the old handler).

And we might want to migrate user settings when some server does a breaking change that requires us to update user's settings.

I think my suggestion for having two separate endpoint is still worth it.

rwols: Well, saving is quite simple with `sublime.save_settings('LSP-{}.sublime-settings'.format(cls.name()))`. And server settings can be obtained with `settings.get("settings")`. One can do the migration in the `def configuration` function. I don't see the problem?

rchl: - The name of the settings file is not explicit so have to be "guessed" now. Granted, it's not hard to guess but still. Maybe it should be explicitly stated by the plugin like before?
- Otherwise I don't have strong arguments against it.

rwols: > The name of the settings file is not explicit so have to be "guessed" now.

You're right, it makes sense to do this a bit more explicit.

rchl:
> Honestly why are we even bothering with all these tiny packages? We could have one large LSP-node-based-language-server-pack that downloads bash/intelephense/css/eslint/html/what-have-you.

Brave idea. :)
Some of those need custom request/notification handlers.
Some prefer settings to be separated (and also for new API it could have benefits, as you said).

rwols:
If we're at the point where the `register_plugin` and `unregister_plugin` functions actually work, then it's totally doable to put all of the node-based plugins in one large py file. Separated sublime-settings files can still exist. Custom request/notification handlers are methods of the AbstractPlugin, so that would also work.

predragnikolic:
I prefer more the cons of having separate LSP-* packages,
than the cons of having the one package that tries to cover everything.

Here are the cons I see of having one huge plugin:
- I see it more difficult to maintain. 
If there is a bug, or a feature need to be added to a particular language server, 
or a language server gets updated. Each time you fix/add/update a thing for a particular language server, you need to make a release. And people not using that particular language server will get the update as well. I see no point in that. 

How is the user suppose to install a particular language server?
I belive it will be something in like this: 
You first need to install this middleware plugin(the huge plugin), and after you can install the language server you want and then you are good to go. 
I find the current way easier. If you want css, just install LSP-css, if you want json, just install it LSP-json... 

Basically I see more problems with the one huge monolithic approach, than I do by having them separated.

rchl:
We don't need this anymore, right?
Or there will be some use for it in the future?

rwols:
Indeed, this is redundant now.

PR #15, opened 2020-06-26
Convert languages array to new syntax to avoid debug warnings

rchl:


rchl:
JS/TS needs `document_selector` (`javascript` vs. `source.js`)

rchl:
But maybe I'm not understanding what you are saying. Do you mean that those didn't really need `document_selector` and it might not be correct to map `scopes` to `document_seleector`?

PR #19, opened 2020-07-14
Prefer new (ST4) language settings over the old ones if both are defined

rchl:
@jfcherng

### DIFF
--- a/st3/lsp_utils/npm_client_handler_v2.py
+++ b/st3/lsp_utils/npm_client_handler_v2.py
@@ -100,7 +100,11 @@ class NpmClientHandler(AbstractPlugin):
     def _upgrade_languages_list(cls, languages):
         upgraded_list = []
         for language in languages:
-            if 'scopes' in language:
+            if 'document_selector' in language:
+                language.pop('scopes', None)
+                language.pop('syntaxes', None)
+                upgraded_list.append(language)
+            elif 'scopes' in language:
                 upgraded_list.append({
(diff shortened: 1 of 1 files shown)


st3/lsp_utils/npm_client_handler_v2.py
```diff
@@ -100,7 +100,11 @@ def configuration(cls) -> Tuple[sublime.Settings, str]:
...
         for language in languages:
-            if 'scopes' in language:
+            if 'document_selector' in language:
+                language.pop('scopes', None)
+                language.pop('syntaxes', None)
+                upgraded_list.append(language)
+            elif 'scopes' in language:
                 upgraded_list.append({
                     'languageId': language.get('languageId'),
                     'document_selector': ' | '.join(language.get('scopes')),
```
rwols, line 110: Like I mentioned in the previous PR that introduced this, using `scopes` for the `document_selector` is wrong and you should use `syntaxes` instead. I think this introduced the bug. Anyway, this probably avoids the problem / solves it in a different way :)

jfcherng:
This PR + https://github.com/sublimelsp/LSP-intelephense/pull/37 work well in my clean Win7 VM test.

st3/lsp_utils/npm_client_handler_v2.py
rchl: You weren't strict enough in stopping me from integrating it. :)

PR #20, opened 2020-08-08
Allow user to specify extra command-line arguments for server binary

rchl:
Resolves https://github.com/sublimelsp/LSP-typescript/issues/23

### DIFF
--- a/st3/lsp_utils/npm_client_handler.py
+++ b/st3/lsp_utils/npm_client_handler.py
@@ -76,12 +76,20 @@ class NpmClientHandler(LanguageHandler):
     def config(self) -> ClientConfig:
         assert self.__server
 
+        client_configuration = self._read_configuration()
+        command = ['node', self.__server.binary_path] + self.get_binary_arguments()
+        extra_args = client_configuration.get('extra_binary_arguments', None)
+        if isinstance(extra_args, list):
+            command += extra_args
+
         configuration = {
             'enabled': True,
-            'command': ['node', self.__server.binary_path] + self.get_binary_arguments(),
+            'command': command,
         }
 
-        configuration.update(self._read_configuration())
+        for key, default in CLIENT_SETTING_KEYS.items():
+            configuration[key] = client_configuration.get(key, default)
+
         self.on_client_configuration_ready(configuration)
         return read_client_config(self.name, configuration)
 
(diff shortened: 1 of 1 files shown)


st3/lsp_utils/npm_client_handler.py
```diff
@@ -76,12 +76,20 @@ def name(self) -> str:
     def config(self) -> ClientConfig:
         assert self.__server
 
+        client_configuration = self._read_configuration()
+        command = ['node', self.__server.binary_path] + self.get_binary_arguments()
+        extra_args = client_configuration.get('extra_binary_arguments', None)
```
predragnikolic, line 81: I guess you need to add `extra_binary_arguments` to the `CLIENT_SETTING_KEYS`,
because `client_configuration` will only contain dict keys for values defined in CLIENT_SETTING_KEYS.

predragnikolic: [Discussion/Question]
How about exposing the `command` in the LSP-*.sublime-setting?
So for example for the LSP-typescript
in LSP-typescript.sublime-settings we will have:
```diff
{
        "command": ["${cache_path}/node_modules/typescript-language-server/lib/cli.js", "--stdio"],
	"languages":[],
	"initializationOptions": {},
	"settings": {},
}
```
That way,
The user can add any additional argument to the server
and we can remove the `get_binary_arguments` method on the interface and potentially the `extra_binary_arguments` client config if this gets merged. But as a downside exposes it to the user so he can mess things up.  

To be honest I haven't thought about this... it was just as an idea that came to my mind, maybe it gives some ideas to you

rchl: > How about exposing the `command` in the LSP-*.sublime-setting?

I was against it before but I think we didn't have `${cache_path}` before. So I think I'm OK with it now.

rchl: Just that when I think of it now, `${cache_path}` is not supported by LSP3 and I don't want to maintain two versions.

I'm actually considering backporting that to LSP3 if possible. And also DottedDict. No idea if it's feasible though...

predragnikolic: > is not supported by LSP3 

I forgot about that.
Backporting those features to ST3 would be good, but a lot had changed between the ST3 and ST4 branch... So i don't know what to say

rchl:
Superseded by #28

PR #23, opened 2020-09-12
Add support for specifying minimum Node version

rchl:
A plugin can define a class method like:

@classmethod
def minimum_node_version(cls) -> Tuple[int, int, int]:
    return (10, 0, 0)

to specify the minimum Node version required by the server. By default
the minimum is v8.

Relates to https://github.com/sublimelsp/LSP-pyright/issues/15

### DIFF
--- a/st3/lsp_utils/npm_client_handler.py
+++ b/st3/lsp_utils/npm_client_handler.py
@@ -5,7 +5,7 @@ from LSP.plugin.core.protocol import Notification
 from LSP.plugin.core.protocol import Request
 from LSP.plugin.core.protocol import Response
 from LSP.plugin.core.settings import ClientConfig, read_client_config
-from LSP.plugin.core.typing import Any, Callable, Dict, Optional
+from LSP.plugin.core.typing import Any, Callable, Dict, Optional, Tuple
 import shutil
 import sublime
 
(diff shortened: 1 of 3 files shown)


rchl:
@AmjadHD

rwols:
Pretty handy. I think we can generalize this to all plugins in some way.

rchl:
I guess pip comes to mind with its "requirements". But I can't easily see a generic solution to handle both.

For this use case we need to:
 - run command (`node --version`)
 - parse the version output
 - compare with minimum version that is hardcoded

How could this be handled in a declarative way? And how would be required for `pip`? And what would be a generic name for of such method?

rwols:
> How could this be handled in a declarative way?

We can add a new method to `AbstractPlugin` that asks for the global requirements. By default it can have no global requirements:

```python
class AbstractRequirement(metaclass=ABCMeta):

    @abstractmethod
    def satisfied(self) -> Optional[str]:
        """Return None if the requirement is satisfied, otherwise return an explanation of the failing requirement"""
        pass


class NodeRequirement(AbstractRequirement):

    def __init__(self, min_version: Tuple[int, int, int]) -> None:
        ...

    def satisfied(self) -> Optional[str]:
        if self.node_is_missing():
            return "node is not in $PATH"
        elif self.node_version_is_too_low():
            return "node required version is at least X.Y.Z. You have version A.B.C"
        else:
            return None


class AbstractPlugin:

    @classmethod
    def global_requirements(cls) -> Optional[List[AbstractRequirement]]:
        return None

class NodeBasedPlugin(AbstractPlugin):

    @classmethod
    def global_requirements(cls) -> Optional[List[AbstractRequirement]]:
        return [NodeRequirement(12, 0, 0)]
```

PR #30, opened 2020-10-01
Add get_session_buffer_for_uri to LSP4 API

rchl:


rchl:
Hmm, I didn't really have to add it here since LSP-eslint has a separate branch for ST4...

st3/lsp_utils/npm_client_handler_v2.py
```diff
@@ -180,3 +181,18 @@ def __init__(self, weaksession: 'weakref.ref[Session]') -> None:
...
+        Return the session buffer for the requested URI.
+
+        :param    uri:  The URI to get session buffer for.
+
+        :returns: An optional session buffer if found for the URI.
+        """
+        session = self.weaksession()
+        if session:
+            return session.get_session_buffer_for_uri_async(uri)
+        return None
```
rwols, line 196: Why not add this to `AbstractPlugin` in LSP?

rchl: As you see I've dropped this from lsp_utils. No need to create duplication as I can just access that public method from Session directly.

PR #31, opened 2020-10-08
Add support for installing servers outside of ST "catalog"

rchl:
A large number of files within directories that ST considers part of the
"catalog" make ST less performant, making catalog queries and ST startup
take a longer time.

Introduced a class method that plugins can override to request being
installed in the "Data/LSP-servers/[PackageName]" directory which is
part of the user settings but not inside directories that are part of
the "catalog".

This needs to be done in an opt-in way instead of just switching directly
in lsp_utils as we need to allow the update flow to run for dependent
packages so that the "plugin_unload" event clears out the old directory.

### DIFF
--- a/st3/lsp_utils/npm_client_handler.py
+++ b/st3/lsp_utils/npm_client_handler.py
@@ -75,7 +75,8 @@ class NpmClientHandler(LanguageHandler):
             cls.__server = get_server_npm_resource_for_package(
                 cls.package_name, cls.server_directory, cls.server_binary_path,
                 cls.package_storage(), cls.minimum_node_version())
-            cls.__server.setup()
+            if cls.__server:
+                cls.__server.setup()
 
     @classmethod
     def cleanup(cls) -> None:
@@ -89,7 +90,7 @@ class NpmClientHandler(LanguageHandler):
     @classmethod
     def additional_variables(cls) -> Optional[Dict[str, str]]:
         return {
-            'server_path': cls.__server.binary_path
+            'server_path': cls.binary_path()
         }
 
     @classmethod
@@ -105,18 +106,20 @@ class NpmClientHandler(LanguageHandler):
         return os.path.join(storage_path, cls.package_name)
 
     @classmethod
+    def binary_path(cls) -> str:
+        return cls.__server.binary_path if cls.__server else ''
+
+    @classmethod
     def install_in_cache(cls) -> bool:
         return True
 
     @property
     def config(self) -> ClientConfig:
-        assert self.__server
-
-        configuration = {'enabled': True}  # type: Dict[str, Any]
+        configuration = {'enabled': self.__server != None}  # type: Dict[str, Any]
         configuration.update(self._read_configuration())
 
         if not configuration['command']:
-            configuration['command'] = ['node', self.__server.binary_path] + self.get_binary_arguments()
+            configuration['command'] = ['node', self.binary_path()] + self.get_binary_arguments()
 
         self.on_client_configuration_ready(configuration)
         base_settings_path = 'Packages/{}/{}'.format(self.package_name, self.settings_filename)
@@ -185,9 +188,7 @@ class NpmClientHandler(LanguageHandler):
         if not is_node_installed():
             sublime.status_message("{}: Please install Node.js for the server to work.".format(cls.package_name))
             return False
-        if not cls.__server:
-            return False
-        return cls.__server.ready
+        return cls.__server != None and cls.__server.ready
 
     def on_initialized(self, client) -> None:
         """
--- a/st3/lsp_utils/npm_client_handler.py
+++ b/st3/lsp_utils/npm_client_handler.py
@@ -72,7 +72,7 @@ class NpmClientHandler(LanguageHandler):
         assert cls.server_binary_path
         if not cls.__server:
             cls.__server = ServerNpmResource(cls.package_name, cls.server_directory, cls.server_binary_path,
-                                             cls.minimum_node_version())
+                                             cls.minimum_node_version(), cls.install_in_cache())
         cls.__server.setup()
 
     @classmethod
@@ -93,6 +93,10 @@ class NpmClientHandler(LanguageHandler):
     @classmethod
     def minimum_node_version(cls) -> Tuple[int, int, int]:
         return (8, 0, 0)
+
+    @classmethod
+    def install_in_cache(cls) -> bool:
+        return True
 
     @property
     def config(self) -> ClientConfig:
--- a/st3/lsp_utils/npm_client_handler.py
+++ b/st3/lsp_utils/npm_client_handler.py
@@ -1,11 +1,12 @@
 from .api_wrapper import ApiWrapperInterface
-from .server_npm_resource import ServerNpmResource
+from .server_npm_resource import get_server_npm_resource_for_package, ServerNpmResource
 from LSP.plugin.core.handlers import LanguageHandler
 from LSP.plugin.core.protocol import Notification
 from LSP.plugin.core.protocol import Request
 from LSP.plugin.core.protocol import Response
 from LSP.plugin.core.settings import ClientConfig, read_client_config
 from LSP.plugin.core.typing import Any, Callable, Dict, Optional, Tuple
+import os
 import shutil
 import sublime
 
@@ -50,11 +51,11 @@ class ApiWrapper(ApiWrapperInterface):
 
 class NpmClientHandler(LanguageHandler):
     # To be overridden by subclass.
-    package_name = None
-    server_directory = None
-    server_binary_path = None
+    package_name = ''
+    server_directory = ''
+    server_binary_path = ''
     # Internal
-    __server = None
+    __server = None  # type: Optional[ServerNpmResource]
 
     def __init__(self):
         super().__init__()
@@ -71,9 +72,10 @@ class NpmClientHandler(LanguageHandler):
         assert cls.server_directory
         assert cls.server_binary_path
         if not cls.__server:
-            cls.__server = ServerNpmResource(cls.package_name, cls.server_directory, cls.server_binary_path,
-                                             cls.minimum_node_version(), cls.install_in_cache())
-        cls.__server.setup()
+            cls.__server = get_server_npm_resource_for_package(
+                cls.package_name, cls.server_directory, cls.server_binary_path,
+                cls.package_storage(), cls.minimum_node_version())
+            cls.__server.setup()
 
     @classmethod
     def cleanup(cls) -> None:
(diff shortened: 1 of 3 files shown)


rchl:
Even though the code for checking and installing dependencies is a bit different, it should work as before by default, without opting-in to the new behavior.

st3/lsp_utils/server_npm_resource.py
```diff
@@ -85,7 +86,11 @@ def setup(self) -> None:
             return
 
         self._initialized = True
-        self._package_cache_path = os.path.join(sublime.cache_path(), self._package_name)
+        if self._install_in_cache:
+            data_dir = sublime.cache_path()
+        else:
+            data_dir = os.path.normpath(os.path.join(sublime.cache_path(), '..'))
+        self._package_cache_path = os.path.join(data_dir, 'LSP-servers', self._package_name)
```
rwols, line 93: Can we call it LSP-Data instead of LSP-servers? It could also be used for log files and such. All other directories in the $DATA dir are also titlecase.

st3/lsp_utils/npm_client_handler.py
```diff
@@ -94,6 +94,10 @@ def additional_variables(cls) -> Optional[Dict[str, str]]:
     def minimum_node_version(cls) -> Tuple[int, int, int]:
         return (8, 0, 0)
 
+    @classmethod
+    def install_in_cache(cls) -> bool:
+        return True
```
rwols, line 99: Let's also add a `basedir()` class method to AbstractPlugin?

rchl: You mean to define where it gets installed?

rwols: Yes :)

rchl: Then the method name should probably be something more descriptive like "install_basedir()".

But I don't really get why should that be on the plugin interface since it's the plugin itself that handles installation and it's nothing that LSP (package) itself is concerned about.

st3/lsp_utils/server_npm_resource.py
```diff
@@ -85,7 +86,11 @@ def setup(self) -> None:
             return
 
         self._initialized = True
-        self._package_cache_path = os.path.join(sublime.cache_path(), self._package_name)
+        if self._install_in_cache:
+            data_dir = sublime.cache_path()
+        else:
+            data_dir = os.path.normpath(os.path.join(sublime.cache_path(), '..'))
+        self._package_cache_path = os.path.join(data_dir, 'LSP Data', self._package_name)
```
rwols, line 93: Missing `-`, or was this intentional?

st3/lsp_utils/npm_client_handler.py
rwols: There are more plugins than just those based on npm.

predragnikolic:
++

Just to drop some ideas, maybe you get an idea from it:

### First thought:
I think we should try to standardize a directory where plugins(not just LSP) are allowed to put stuff in in(knowing that that folder will not be part of the sublime catalog, and wont slow down ST). The name of that folder could be:
* `Storage`
* `Plugin Storage` - (ST uses a space, so it would be ok to be consent, example  `Installed Packages`)
* Something else...

So right next to `Installed Packages`, `Packages`, `Cache`, there will be the STORAGE folder where plugins are allowed to put data.

### Second thought: 

if we add a folder like that, it would be good if we create a function similar to `cache_path()`, `packages_path()`
called `storage_path()` which would return that path. 


### Third thought:
Maybe there is no need to define a method `install_in_cache` on the interface. Maybe I am wrong here, but can we get away without adding the `install_in_cache` method nor on the abstract plugin nor on the npm langauge handler? :)

st3/lsp_utils/server_npm_resource.py
rchl: Wanted to be consistent with `Installed Packages`.

st3/lsp_utils/npm_client_handler.py
rchl: Yes but why would that path be defined on the AbstractPlugin interface. The installation is entirely handled by the plugin (regardless if npm-based or not).

rchl:
> I think we should try to standardize a directory where plugins(not just LSP) are allowed to put stuff in in(knowing that that folder will not be part of the sublime catalog, and wont slow down ST). The name of that folder could be:

How do we go about it and who will be the one to decide that? :)

> if we add a folder like that, it would be good if we create a function similar to `cache_path()`, `packages_path()`
> called `storage_path()` which would return that path.

Who should create that function? I think it only makes sense that ST natively does that in that case.

> Maybe there is no need to define a method `install_in_cache` on the interface. Maybe I am wrong here, but can we get away without adding the `install_in_cache` method nor on the abstract plugin nor on the npm langauge handler? :)

If you are thinking about just straight changing the path here then read the PR description.

st3/lsp_utils/server_npm_resource.py
rwols: Ok :) I actually like the space.

predragnikolic:
for
> How do we go about it and who will be the one to decide that? :)
and
> Who should create that function? I think it only makes sense that ST natively does that in that case.

I could ask on Discord for feedback on the plugin development channel, 
I would like to see if you would find something like that useful. (to see if it is worth putting effort into something like that).

I think that it would be ok if ST offered a `sublime.storage_path()` where plugins can store data, without affecting ST performance.

rchl:
If you want to lead that effort then I will support you but at the same time, I want to be pragmatic and get s*** done. ;)

rchl:
Will put some performance info related to the Discord discussion as it will be lost there soon:

With ~90k extra files in Cache, startup time, as reported in the console takes from 2.2 to 2.6s. Without those extra files it takes about 0.9s.
With those extra files, querying for all resources takes ~320ms, without just 50ms.
Used this code to query for resources:

```py
import time; start = time.time(); r = sublime.find_resources('*'); int(round((time.time() - start) * 1000))
```

(not sure what takes more time - querying or allocating the memory for the results)

st3/lsp_utils/server_npm_resource.py
```diff
@@ -85,7 +86,11 @@ def setup(self) -> None:
             return
 
         self._initialized = True
-        self._package_cache_path = os.path.join(sublime.cache_path(), self._package_name)
+        if self._install_in_cache:
+            self._package_cache_path = os.path.join(sublime.cache_path(), self._package_name)
+        else:
+            data_dir = os.path.normpath(os.path.join(sublime.cache_path(), '..'))
+            self._package_cache_path = os.path.join(data_dir, 'Package Storage', self._package_name)
```
rwols, line 93: This doesn't account for the node.js version installed on the machine.

predragnikolic: No it doesn't, but I like how simple it is without having different folders for different node versions. For this case i would choose simplicity over complexity

rwols: You can ignore the complexity, but the complexity won't ignore you :)

Some user will eventually pop up and say "yes, I have a portable ST install on my USB stick and my USB stick's flash memory is now dead, because I switch machines all the time and these node modules are constantly being overwritten".

predragnikolic: I am not saying that we should ignore it, rather wait until we maybe come up with maybe something simpler.

Here is why I think it is complex. If we going to have a folder for each node version than that solution doesn't scale, because there are a lot of node versions out there. (We may end up with at least 6 folders)

We can open an issue for that and discuss the implementation there. (Plus I don't think that node versions are the concern of this PR)

rwols: > We can open an issue for that and discuss the implementation there. (Plus I don't think that node versions are the concern of this PR)

A fair point. However, given the desire of Rafal to keep everything backwards compatible, you would have to write yet more confusing migration code in another pull request. So I propose to do it right away.

rchl: I'll add that when I have some free time.

But I suppose it will have to work like so:
 - user installs the package and triggers installation within node v12, for example
 - user updates node to v14 and starts ST - v14 of the server gets installed alongside the v12 server

So in that scenario, the v12 version would not be removed until the user uninstalls the package, even if user is never gonna go back to v12 of node.

rchl: It would need some refactoring to get the node version before checking if dependencies are installed... Also, that means an extra subprocess start every time the plugin is registered.

st3/lsp_utils/npm_client_handler.py
jfcherng: I think it's good to be `False` by default?

rchl: It will be false once all existing packages are migrated to new structure. For now it's opt in so that on upgrading lsp-* packages, the old directory is cleaned up.

But now I'm wondering if upgrading dependency is maybe disabling all packages that depend on it... In that case we could set it to false straight away.

st3/lsp_utils/server_npm_resource.py
rchl: OK. Doing node version check now before checking if server is installed.

That means that it happens on the main thread but I'll move that to `AbstractPlugin.install_or_update` later for ST4.

I'm only doing the check once per session through a new `NodeVersionResolver` singleton.

Also, I've discovered that in some cases the `ServerNpmResource` would start installing dependencies simultaneously for two or more instances of AbstractPlugin. That happened when ST was started with multiple windows open with each triggering the same LSP plugin. So I've added a `get_server_npm_resource_for_package` utility function that returns an instance of `ServerNpmResource` that is keyed to specific server paths so that multiple instances of the same LSP plugin share the same instance.

st3/lsp_utils/npm_client_handler.py
rchl: I've asked on Discord and was told that updating dependency doesn't disable all packages that depend on it so unfortunately, we need to handle that per LSP package.

st3/lsp_utils/server_npm_resource.py
rwols: > That happened when ST was started with multiple windows open with each triggering the same LSP plugin.

Hmm... I don't think it's going to be solvable in LSP lest we end up in callback hell. Oh well. Let's wait till we can upgrade to python 3.8 :-)

rchl: I think that might an issue specific to ServerNpmResource which starts a new thread for installing node dependencies.

When using `AbstractPlugin.install_or_update` it should run sequentially since LSP will trigger both from the same thread so only first one should trigger installation.

I'm planning to switch to `AbstractPlugin.install_or_update` in next PR. Right now `ServerNpmResource` is triggered from `plugin_loaded`.

PR #32, opened 2020-10-18
Add VSCode marketplace client handler

jfcherng:
There are some LSP servers, which are only available either on VSCode's marketplace or as `.vsix` files.

- TailwindCSS: https://github.com/tailwindlabs/tailwindcss-intellisense
  which was asked on https://github.com/sublimelsp/LSP/issues/1377#issuecomment-711036812
- Roblox: https://marketplace.visualstudio.com/items?itemName=Nightrains.robloxlsp
  which was asked on https://discordapp.com/channels/280102180189634562/645268178397560865/767193838379859969

This PR proposes a dedicated handler, which should simplify `LSP-*` plugin codes for them.

It

- downloads the extension from the marketplace with given extension `itemName` and `version`
- unzips the downloaded `.vsix` file
- checks `node` executable availability if necessary

---

An example use for https://marketplace.visualstudio.com/items?itemName=Nightrains.robloxlsp

<details>

```py
import os
import sublime

from LSP.plugin.core.typing import List
from lsp_utils import VscodeMarketplaceClientHandler


def get_server_bin_path(platform: str = sublime.platform()) -> str:
    """
    @brief Get the LSP server binary path.
    @param platform The platform ("osx", "linux" or "windows")
    @return The LSP server binary path.
    """

    os_bin = {
        "linux": ["Linux", "lua-language-server"],
        "osx": ["macOS", "lua-language-server"],
        "windows": ["Windows", "lua-language-server.exe"],
    }

    return os.path.join("extension", "server", "bin", *(os_bin[platform]))


class LspRobloxPlugin(VscodeMarketplaceClientHandler):
    package_name = "LSP-Roblox"

    # @see https://marketplace.visualstudio.com/items?itemName=Nightrains.robloxlsp
    extension_item_name = "Nightrains.robloxlsp"
    extension_version = "0.11.5"
    server_binary_path = get_server_bin_path()
    execute_with_node = False

    @classmethod
    def get_binary_arguments(cls) -> List[str]:
        basename = "{}.sublime-settings".format(cls.package_name)
        settings = sublime.load_settings(basename)

        return [
            "-E",
            os.path.join(
                "${package_cache_path}",
                "extension",
                "server",
                settings.get("main_lua_file"),
            ),
        ]
```

</details>

### DIFF
--- a/st3/lsp_utils/__init__.py
+++ b/st3/lsp_utils/__init__.py
@@ -5,14 +5,19 @@ except ImportError:
 
 from .api_wrapper import ApiWrapperInterface
 from .server_npm_resource import ServerNpmResource
+from .server_vscode_marketplace_resource import ServerVscodeMarketplaceResource
 
 if lsp_version >= (1, 0, 0):
     from .npm_client_handler_v2 import NpmClientHandler
+    from .vscode_marketplace_client_handler_v2 import VscodeMarketplaceClientHandler
 else:
     from .npm_client_handler import NpmClientHandler
+    from .vscode_marketplace_client_handler import VscodeMarketplaceClientHandler
 
 __all__ = [
     'ApiWrapperInterface',
     'NpmClientHandler',
     'ServerNpmResource',
+    'VscodeMarketplaceClientHandler',
+    'ServerVscodeMarketplaceResource',
 ]
--- a/st3/lsp_utils/server_vscode_marketplace_resource.py
+++ b/st3/lsp_utils/server_vscode_marketplace_resource.py
@@ -6,6 +6,13 @@ import sublime
 import urllib.request
 import zipfile
 
+try:
+    from LSP.plugin.core.views import get_storage_path
+except ImportError:
+    # polyfill for LSP in ST 3
+    def get_storage_path() -> str:
+        return os.path.abspath(os.path.join(sublime.cache_path(), "..", "Package Storage"))
+
 
 def log_and_show_message(msg, additional_logs: str = None, show_in_status: bool = True) -> None:
     print(msg, "\n", additional_logs) if additional_logs else print(msg)
@@ -32,6 +39,7 @@ class ServerVscodeMarketplaceResource(object):
         extension_item_name: str,
         extension_version: str,
         server_binary_path: str,
+        install_in_cache: bool,
     ) -> None:
         self._initialized = False
         self._is_ready = False
@@ -39,6 +47,7 @@ class ServerVscodeMarketplaceResource(object):
         self._extension_item_name = extension_item_name
         self._extension_version = extension_version
         self._binary_path = server_binary_path
+        self._install_in_cache = install_in_cache
         self._package_cache_path = ""
         self._activity_indicator = None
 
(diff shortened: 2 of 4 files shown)


rwols:
Javascript files downloaded from that "marketplace" probably expect the node.js runtime to be a certain version, namely the one shipped with VSCode. How would you handle this?

jfcherng:
> Javascript files downloaded from that "marketplace" probably expect the node.js runtime to be a certain version, namely the one shipped with VSCode. How would you handle this?

what if something like https://github.com/sublimelsp/lsp_utils/blob/d63ccf051ec39456f8fabe497906b8b6630d3660/st3/lsp_utils/npm_client_handler_v2.py#L85-L87, such as

```py
    @classmethod
    def satisfy_node_version(cls, version: Tuple[int, int, int]) -> bool:
        # if the node version should >= 8.0.0
        return version >= (8, 0, 0)
```

rwols:
I was playing around with general prerequisites in this branch actually: https://github.com/sublimelsp/LSP/commit/a78e9582a1bd638a90475722ba4bf069a1195b24

rwols:
While it's an interesting approach to add general routines for downloading things from the vscode marketplace, in the case of LSP-roblox I'm not totally convinced it's the most clean result. For instance, that vscode extension bundles binary files for all three platforms inside of the extension.

I made an LSP-lua repo where I actually separate out those three binaries into the three platforms, and also remove all unneeded files like debugger.lua and all beta Lua files: https://github.com/sublimelsp/LSP-lua

If you look at the releases page there, each (automated) release has zip assets for all three ST platforms (well, only x64 versions). So the binaries are placed in the package itself and there's no need to download files after the package is installed with package control at all.

There is also a person already waiting to release an LSP-roblox package: https://github.com/sublimelsp/repository/pull/2

jfcherng:
Agree on "if we can have dedicated releases for each platform, it's ideal". The big bundle from the marketplace is for the sake of laziness :laughing:

predragnikolic:
> Javascript files downloaded from that "marketplace" probably expect the node.js runtime to be a certain version, namely the one shipped with VSCode. How would you handle this?

I been looking at different ways to have different node versions locally. 
There is [NVM](https://github.com/nvm-sh/nvm) (Node Version Manager package), but for that to work, the user should have nvm globally installed, which is a deal breaker for me.

Another option is to use `npx`, it is already shipped with `node`.
You can read what `npx` can do, but among other things,
it allows running the script with the specified node version.

For example: 
```
node ./script.js // will execute the script with the users globally installed node version
npx node@6 ./script.js // will execute the script with node version 6
```

here is an example
```
λ  ~/Documents/example  echo "console.log('you')" > hello.js
λ  ~/Documents/example  ls
hello.js
λ  ~/Documents/example  cat hello.js 
console.log('you')
λ  ~/Documents/example  node -v
v12.18.3
λ  ~/Documents/example  node ./hello.js         // executes instantly
you
λ  ~/Documents/example  npx node@6 ./hello.js   // takes a while because `npx node@6` is executed for the first time and it needs to download things from the internet
you
λ  ~/Documents/example  npx node@6 ./hello.js   // it doesn't execute instantly, there is a small delay, but it is way faster than the first time `npx node@6` was called.
you
```

rchl:
I think I have nothing against the idea but there is an awful lot of duplication between existing and new handlers. I think it would make sense to extract common case into a common base.

rchl:
Also, I would like to complete the move to a new server directory before introducing those new handlers.

jfcherng:
Close this due to lack of use case and... I won't have much time for this recently :/

rchl:
I'm not certain if it would be useful either and also wonder about the legality of doing that but I'm working on some refactorings in lsp_utils that will make it much easier to implement new handlers without duplication.

PR #33, opened 2020-10-25
Make ST4 handler use AbstractPlugin installation flow

rchl:
Instead of creating extra thread when installing server in ST4,
use "AbstractPlugin.needs_update_or_installation" and co.

The advantages are:
 - no extra threads
 - lazy installation - only starts when associated file is opened
 - does not require ST restart after server was installed

 Also, call "(un)register_plugin" as recommended by AbstractPlugin API.

### DIFF
--- a/st3/lsp_utils/helpers.py
+++ b/st3/lsp_utils/helpers.py
@@ -1,0 +1,58 @@
+from LSP.plugin.core.typing import Callable, List, Optional, Tuple
+import re
+import sublime
+import subprocess
+import threading
+
+StringCallback = Callable[[str], None]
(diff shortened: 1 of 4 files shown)


rchl:
I'm ready to merge this and will do so tomorrow if nobody stops me.

rwols:
I only glanced over it :) but it looks like this will work as expected from what I can tell.

rchl:
I've realized now that due to making server initialization lazy, the cleanup that would happen on `plugin_unloaded` (on upgrading LSP-* packages) will likely not happen... So that kinda defeats the purpose of having `install_in_cache` flag and updating packages one-by-one.

PR #37, opened 2020-11-10
Refactor to allow for easier creation of additional handlers

rchl:
Resolves https://github.com/sublimelsp/LSP-json/issues/72


rchl:
Generated documentation: https://sublimelsp.github.io/lsp_utils/

st3/lsp_utils/server_resource_interface.py
```diff
@@ -0,0 +1,69 @@
...
+
+    UNINITIALIZED = 1
+    """Initial status of the server."""
+    ERROR = 2
+    """Initiallation or update has failed."""
+    READY = 3
+    """Server is ready to provide resources."""
+
+
+class ServerResourceInterface(metaclass=ABCMeta):
```
rchl, line 23: Should add some generic implementation that reports installation progress in ST3.

rwols:
I'm seeing this warnings, is it of concern?

```
~/L/A/S/P/l/docs (feat/refactor|✔) $ make html
sphinx-build -M html "source" "."
Sphinx v3.3.0 start op
aanmaken doelmap... klaar
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 2 source files that are out of date
updating environment: [new config] 2 added, 0 changed, 0 removed
reading sources... [100%] index
WARNING: autodoc: failed to import class 'GenericClientHandler' from module 'lsp_utils'; the following exception was raised:
No module named 'LSP'
WARNING: autodoc: failed to import class 'NpmClientHandler' from module 'lsp_utils'; the following exception was raised:
No module named 'LSP'
WARNING: autodoc: failed to import class 'ServerStatus' from module 'lsp_utils'; the following exception was raised:
No module named 'LSP'
WARNING: autodoc: failed to import class 'ServerResourceInterface' from module 'lsp_utils'; the following exception was raised:
No module named 'LSP'
WARNING: autodoc: failed to import class 'ApiWrapperInterface' from module 'lsp_utils'; the following exception was raised:
No module named 'LSP'
WARNING: autodoc: failed to import module 'helpers' from module 'lsp_utils'; the following exception was raised:
No module named 'LSP'
looking for now-outdated files... none found
pickling environment... klaar
checking consistency... klaar
preparing documents... klaar
writing output... [100%] index
generating indices... klaar
writing additional pages... search klaar
copying static files... klaar
copying extra files... klaar
dumping search index in English (code: en)... klaar
dumping object inventory... klaar
build gelukt, 6 warnings.
```

rchl:
See this part: https://github.com/sublimelsp/lsp_utils/blob/5765e870622c4a246d5d6e001c456d1f5a1eca35/docs/source/conf.py#L17-L25

Basically, sphinx has to be told where LSP package and various dependencies are. Same as we just did for LSP-pyright.

I suppose I will have to automate docs deployment through github actions where it will have to clone all those dependencies. Actually can do that locally from makefile first...

docs/Makefile
```diff
@@ -8,6 +8,9 @@ BUILDDIR      = .
 
 install:
 	pip3 install sphinx sphinx-rtd-theme sphinx-autodoc-typehints ghp-import
+	rm -rf source/modules
+	mkdir -p source/modules
+	cd source/modules && git clone --branch st4000-exploration --depth 1 -- git@github.com:sublimelsp/LSP.git
```
rwols: Can't we just ask the user to supply the path to the packages dir?

rchl: I wanted it to be contained and work universally, also on CI. Why complicate it? What benefit would it have to use your own packages?

rwols:
I don't really like how the build artifacts are now committed into the git history in a gh-pages branch to be honest. Is it possible that a github.io website can also be set up in such a way that it downloads an artifact from a "release" ?

rchl:
> I don't really like how the build artifacts are now committed into the git history in a gh-pages branch to be honest. Is it possible that a github.io website can also be set up in such a way that it downloads an artifact from a "release" ?

I think github pages can only serve the content of the branch.
How that content ends up here is another matter. I could configure hook to only publish from tagged commits. Otherwise I'm not sure what's the concern.

More info at https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages

rwols:
I see. So committing build artifacts is the status quo :) it pollutes the git repo with files that nobody will look at :( oh well, if this practice is even in the official docs I guess it's okay.

rchl:
> it pollutes the git repo with files that nobody will look at

If that eases the paint a little, the changes are force-pushed so old changes eventually disappear from the repo (are garbage collected) ;)

st3/lsp_utils/server_resource_interface.py
rchl: That was handled in the client handler actually so I don't have a use case for adding a generic server resource implementation at the moment.

st3/lsp_utils/_client_handler/abstract_plugin.py
```diff
@@ -0,0 +1,161 @@
+from ..api_wrapper_interface import ApiWrapperInterface
+from ..server_resource_interface import ServerStatus
+from .interface import ClientHandlerInterface
+from LSP.plugin import AbstractPlugin
+from LSP.plugin import ClientConfig
+from LSP.plugin import Notification
```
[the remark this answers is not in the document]
rchl: I think there should be no problems with that.

st3/lsp_utils/_client_handler/__init__.py
```diff
@@ -0,0 +1,15 @@
+from LSP.plugin import __version__ as lsp_version
+
+
+major_version = lsp_version[0]
+
+if major_version == 1:  # types are problematic when comparing tuples with ">=".
```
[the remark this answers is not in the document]
rchl: Python yes but both sphinx and pyright were complaining.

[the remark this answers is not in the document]
rchl: No, I lied, Sphinx seems to be fine. It's only pyright:

Screenshot 2020-11-13 at 22 19 33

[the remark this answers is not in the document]
rwols: I noticed this too when working on LSP-metals.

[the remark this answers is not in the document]
rchl: https://github.com/microsoft/pyright/issues/1170

st3/lsp_utils/_client_handler/abstract_plugin.py
```diff
@@ -0,0 +1,161 @@
...
+    @classmethod
+    def can_start(cls, window: sublime.Window, initiating_view: sublime.View,
+                  workspace_folders: List[WorkspaceFolder], configuration: ClientConfig) -> Optional[str]:
+        if cls.manages_server():
+            server = cls.get_server()
+            if not server or server.get_status() == ServerStatus.ERROR:
+                return "{}: Error installing server dependencies.".format(cls.package_name)
+            if server.get_status() != ServerStatus.READY:
+                return "{}: Server installation in progress...".format(cls.package_name)
+        message = cls.is_allowed_to_start(window, initiating_view, workspace_folders, configuration)
```
rwols, line 96: While trying to fix https://github.com/sublimelsp/LSP/issues/1440, I learned that if you listen for a change in a sublime.Settings object, then your plugin should never use `.set()` on that object as it can get into an infinite recursion.

Currently I'm guilty of doing that in a bunch of my helper packages. I will have to change that. I think the best place for these kind of changes is actually here in `can_start` (or `is_allowed_to_start` in this interface), and to do the mutation on the ClientConfig object.

Unfortunately LSP-pyright now also mutates its sublime.Settings object to dynamically change the extraPaths. What's also unfortunate is that the old `on_start` doesn't receive enough information.

st3/lsp_utils/_client_handler/language_handler.py
```diff
@@ -0,0 +1,146 @@
...
+        if not settings_dict['command']:
+            settings_dict['command'] = self.get_command()
+        return read_client_config(self.name, settings_dict, filepath)
+
+    @classmethod
+    def on_start(cls, window: sublime.Window) -> bool:
+        if cls.manages_server():
+            server = cls.get_server()
+            return server is not None and server.get_status() == ServerStatus.READY
+        message = cls.is_allowed_to_start(window)
```
rwols, line 78: Ugh, so the old LanguageHandler doesn't receive the ClientConfig?

st3/lsp_utils/_client_handler/abstract_plugin.py
jfcherng: > Unfortunately LSP-pyright now also mutates its sublime.Settings object to dynamically change the extraPaths. 

what if revise configuration in `on_client_configuration_ready`? I do it there originally.

rchl: > While trying to fix [sublimelsp/LSP#1440](https://github.com/sublimelsp/LSP/issues/1440), I learned that if you listen for a change in a sublime.Settings object, then your plugin should never use `.set()` on that object as it can get into an infinite recursion.

That doesn't sound right. I guess the more precise way to say it would be "you can get into an infinite loop if you modify your settings in response to the change listener. I don't currently understand in what circumstances that would happen.

rchl: > > Unfortunately LSP-pyright now also mutates its sublime.Settings object to dynamically change the extraPaths.
> 
> what if revise configuration in `on_client_configuration_ready`? I do it there originally.

Before we ask that question, maybe we should first ask what is the expected flow on the user changing some plugin setting.

Current settings reading is done once at plugin init so whether we use `on_client_configuration_ready` or `on_settings_read` it will not work correctly as those won't re-run on changing settings.

st3/lsp_utils/_client_handler/language_handler.py
rchl: Maybe we could store the ClientConfig that we are creating in `config()` and then we could pass it here. Just not sure whether that could end up with the one that LSP has and the one that LSP-* has, getting out of sync (if LSP would re-create its own).

rwols: LSP always creates a copy in WindowConfigManager.update

st3/lsp_utils/_client_handler/abstract_plugin.py
rwols: Well, they can be re-run provided that LSP calls `plugin.configuration()` on each setting change, but that causes the infinite recursion for some plugins.

So we really shouldn't be using sublime.Settings.set anywhere. So the only place (I think?) to modify things is in ClientConfig, which is passed in `can_start`.

st3/lsp_utils/_client_handler/language_handler.py
rchl: Then I guess we would have to add something on the master branch for that? But the old API was historically hard to extend so not sure how that would work out.

st3/lsp_utils/_client_handler/abstract_plugin.py
rwols: So something like this for pyright could work:

```python
    @classmethod
    def is_allowed_to_start(cls, window, initiating_view, workspace_folders, configuration):
        if configuration and settings.get("dev_environment") == "sublime_text":
            paths = configuration.settings.get("python.analysis.extraPaths")
            if not isinstance(paths, list):
                paths = []
            paths.extend(cls.find_package_dependency_dirs())
            configuration.settings.set("python.analysis.extraPaths", paths)
```

this way we don't change a sublime.Settings object but we still modify the server settings (just in time).

rwols: but it would only work for the ST4 branch...

rwols: although you can add a version check in `on_settings_read`: `if sublime.version() < "4000": ...`, so that would make it work for both ST3 and ST4 :)

rchl: > So we really shouldn't be using sublime.Settings.set anywhere. So the only place (I think?) to modify things is in ClientConfig, which is passed in `can_start`.

There was a use case where we actually wanted to change the settings and even save them to disk. It was for migrating old language settings to the new one.

> So something like this for pyright could work:

This class method will be called multiple times though (once per window at least). Would the plugin have to make sure that it doesn't add redundant items (when extending array) or "configuration" be pristine on each call?

BTW. What about keeping current semantics where the client is allowed to modify and save settings and adding some extra API to be called when settings change?

BTW2. As I've said before, the semantics of `can_start` doesn't really make it clear that you are allowed to modify the client config from it. If we are changing API, maybe there should be a new method added specifically for the purpose of allowing the client to mutate the config.

rwols: > BTW. What about keeping current semantics where the client is allowed to modify and save settings and adding some extra API to be called when settings change?

Agreed.

> BTW2. As I've said before, the semantics of can_start doesn't really make it clear that you are allowed to modify the client config from it. If we are changing API, maybe there should be a new method added specifically for the purpose of allowing the client to mutate the config.

Agreed. I'll try to work on this.

st3/lsp_utils/_client_handler/__init__.py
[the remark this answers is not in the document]
rchl: It was fixed in pyright. :)

st3/lsp_utils/_client_handler/abstract_plugin.py
rchl: I don't think we have arrived at the correct solution yet in LSP.

Either we merge this refactor as is or wait until we figure out something better for LSP.

But it's really a separate issue, I would say.

PR #38, opened 2020-11-15
feat: add decorators for api handlers

jfcherng:
Just sharing an idea I got during playing with a ~~mysterious~~ Python language server + `vs_marketplace_client_handler`.

---

With this PR, you can register message handlers with decorators.

For example,

```py
def on_ready(self, api) -> None:
    api.on_notification('eslint/status', self.handle_status)
    api.on_request('eslint/openDoc', self.handle_open_doc)

    # sometimes a handler handles multiple message methods...
    api.on_notification('intelephense/IndexingStarted', self.handle_indexing)
    api.on_notification('intelephense/IndexingEnded', self.handle_indexing)

def handle_status(self, params) -> None:
    ...

def handle_open_doc(self, params, respond) -> None:
    ...

def handle_indexing(self, params) -> None:
    ...
```

can be written as

```py
from lsp_utils import notification_handler, request_handler

@notification_handler("eslint/status")
def handle_status(self, params) -> None:
    ...
    
@request_handler("eslint/openDoc")
def handle_open_doc(self, params, respond) -> None:
    ...

@notification_handler(["intelephense/IndexingStarted", "intelephense/IndexingEnded"])
def handle_indexing(self, params) -> None:
    ...
```


rchl:
Interesting

Pros:
 - pretty nice and clean syntax
 - avoids the "register" boilerplate

Cons:
 - adds complexity in the code
 - does not necessarily address any issue or annoyance since the current solution is already pretty OK IMO (certainly better than the magic methods in the AbstractPlugin, no offense ;)).

Also, I wonder whether it is types-friendly (haven't tried it myself)? Will it complain if the decorated function doesn't match the expected signature?

rchl:
> ```python
> @as_notification_handler("eslint/status")
> def handle_status(self, params) -> None:
>     pass
> ```

You would have to import the `as_notification_handler` decorator first, right? How would you do that?

jfcherng:
> > ```python
> > @as_notification_handler("eslint/status")
> > def handle_status(self, params) -> None:
> >     pass
> > ```
> 
> You would have to import the `as_notification_handler` decorator first, right? How would you do that?

Ah, yes, the `ApiWrapper` has to be exposed in the `__init__.py` (or put those decorator in something else publicly). Something like

```py

if lsp_version >= (1, 0, 0):  # type: ignore
    # ST 4
    from .vs_marketplace_client_handler_v2 import ApiWrapper
    from .vs_marketplace_client_handler_v2 import VsMarketplaceClientHandler
else:
    # ST 3
    from .vs_marketplace_client_handler import ApiWrapper
    from .vs_marketplace_client_handler import VsMarketplaceClientHandler

__all__ = [
    "ApiWrapper",
    "ApiWrapperInterface",
    "ServerVsMarketplaceResource",
    "VsMarketplaceClientHandler",
]
```

And then you can use it like

```py
from lsp_utils import ApiWrapper

@ApiWrapper.as_notification_handler("eslint/status")
def handle_status(self, params) -> None:
    pass
```

st3/lsp_utils/npm_client_handler.py
```diff
@@ -44,6 +59,26 @@ def send_request(self, method: str, params: Any, handler: Callable[[Optional[str
         self.__client.send_request(
             Request(method, params), lambda result: handler(result, False), lambda result: handler(result, True))
 
+    @staticmethod
+    def as_notification_handler(event_name: str) -> Callable[[_T], _T]:
+        """ Marks the decorated function as a handler for the notification event. """
+
+        def decorator(func: _T) -> _T:
+            setattr(func, ApiWrapper.HANDLER_MARK_NOTIFICATION, event_name)
```
jfcherng, line 67: Should put the registered event_name into a list, otherwise if this decorator is used twice on a same function, only the later one works. But, well, just sharing an idea here :D

jfcherng:
It resolves nothing in practice. Somehow feel it can be useful in some ways (not necessary in this PR).

rchl:
> Ah, yes, the `ApiWrapper` has to be exposed in the `__init__.py` (or put those decorators something else publicly). Something like

Conceptually that would be wrong as it's an implementation-specific class that is not supposed to be exposed.

st3/lsp_utils/npm_client_handler.py
```diff
@@ -172,7 +207,19 @@ def on_initialized(self, client) -> None:
...
+
+    def _register_custom_server_event_handlers(self, api: ApiWrapper) -> None:
+        for _, func in inspect.getmembers(self, predicate=inspect.isroutine):
+            event_name = getattr(func, ApiWrapper.HANDLER_MARK_NOTIFICATION, None)  # type: Optional[str]
+            if event_name:
+                api.on_notification(event_name, func)
+
+            event_name = getattr(func, ApiWrapper.HANDLER_MARK_REQUEST, None)  # type: Optional[str]
+            if event_name:
+                api.on_request(event_name, func)
```
rchl, line 225: I suppose you wouldn't want to decorate the same function with both decorators as that wouldn't work. So should break early.

jfcherng:
> Also, I wonder whether it is types-friendly (haven't tried it myself)? Will it complain if the decorated function doesn't match the expected signature?

On my side, LSP-py* complains after I make the function signature clear.

```py
T_NOTIFICATION_HANDLER = Callable[["VsMarketplaceClientHandler", Dict[str, Any]], None]
```

and

```py
    @staticmethod
    def as_notification_handler(event_name: str) -> Callable[[T_NOTIFICATION_HANDLER], T_NOTIFICATION_HANDLER]:
        """ Marks the decorated function as a handler for the notification event. """

        def decorator(func: T_NOTIFICATION_HANDLER) -> T_NOTIFICATION_HANDLER:
            setattr(func, ApiWrapper.HANDLER_MARK_NOTIFICATION, event_name)
            return func

        return decorator
```

jfcherng:
> > Ah, yes, the `ApiWrapper` has to be exposed in the `__init__.py` (or put those decorators something else publicly). Something like
> 
> Conceptually that would be wrong as it's an implementation-specific class that is not supposed to be exposed.

Or probably put decorators under `NpmClientHandler`, and use `@self.as_notification_handler(...)`.

st3/lsp_utils/__init__.py
```diff
@@ -1,4 +1,5 @@
 from ._client_handler import ClientHandler
+from ._client_handler import as_notification_handler, as_request_handler
```
rchl, line 2: I'm now used to the convention where each import is separate (apart from types).

st3/lsp_utils/_client_handler/decorator.py
```diff
@@ -0,0 +1,39 @@
...
+T_HANDLER = Callable[[Any, Any], None]
+T_SERVER_EVENTS = Union[str, Iterable[str]]
+
+HANDLER_MARKS = {
+    "notification": "__handle_notification_events",
+    "request": "__handle_request_events",
+}
+
+
+def as_notification_handler(server_events: T_SERVER_EVENTS) -> Callable[[T_HANDLER], T_HANDLER]:
```
rchl, line 19: I think the `as_` prefix is redundant. A decorator like `@notification_handler` is as clear and probably more conventional (built-in python decorators like `@classmethod` don't use the `as_` prefix).

st3/lsp_utils/generic_client_handler.py
```diff
@@ -187,4 +188,25 @@ def on_ready(self, api: ApiWrapperInterface) -> None:
 
         :param api: The API instance for interacting with the server.
         """
-        pass
+        self._register_custom_server_event_handlers(api)
+
+    def _register_custom_server_event_handlers(self, api: ApiWrapperInterface) -> None:
```
rchl: `_register_decorated_handlers` maybe?

st3/lsp_utils/generic_client_handler.py
```diff
@@ -187,4 +188,25 @@ def on_ready(self, api: ApiWrapperInterface) -> None:
...
+        :param api: The API instance for interacting with the server.
+        """
+        for _, func in inspect.getmembers(self, predicate=inspect.isroutine):
+            # client_event is like "notification", "request"
+            for client_event, handler_mark in HANDLER_MARKS.items():
+                event_registrator = getattr(api, "on_" + client_event, None)
+                if callable(event_registrator):
+                    server_events = getattr(func, handler_mark, [])  # type: List[str]
+                    for server_event in server_events:
+                        event_registrator(server_event, func)
```
rchl: Does this allow for both decorators to be applied to the same function? That wouldn't make sense as the function signatures differ.

rchl (commented):
You should avoid force-push when there are reviewers involved since that makes it harder to see what has changed.

In the end, when ready to merge, there is an option to squash and merge so the history will get cleaned up.

st3/lsp_utils/_client_handler/language_handler.py
```diff
@@ -144,3 +148,24 @@ def __init__(self):
         # Will be a no-op if already ran.
         # See https://github.com/sublimelsp/LSP/issues/899
         self.setup()
+
+    def _register_custom_server_event_handlers(self, api: ApiWrapperInterface) -> None:
```
rchl, line 152: Does it need to be done separately for each handler implementation. Did the previous approach of changing generic client handler didn't work?

st3/lsp_utils/_client_handler/decorator.py
jfcherng: I am fine with whatever names :)

st3/lsp_utils/_client_handler/language_handler.py
jfcherng: It works but potentially error-prone if people forget to do `super().on_ready()` in their custom `on_ready()` then `_register_custom_server_event_handlers()` wont be executed.

st3/lsp_utils/generic_client_handler.py
jfcherng: yeah... at this moment I just assume those two are all `Callable[[Any, Any], None]`...

st3/lsp_utils/_client_handler/language_handler.py
rchl: But on the other-hand such duplication is worse. The big goal of the refactor was to get rid of as much duplication as possible.

You can add some internal functions to the client interface that will be called by handler implementations before `on_ready` and which generic handler can decorate from.

st3/lsp_utils/generic_client_handler.py
rchl: I would really like the types to be correct and checked so that when decorating with `@notification_handler`, for example, would raise a type error if the function doesn't have a correct number of arguments (and maybe also correct types).

I'm not certain if it's possible for decorators to affect type checking though...

rchl: > yeah... at this moment I just assume those two are all `Callable[[Any, Any], None]`...

Types being incorrect is one thing but I guess it would also crash at runtime since the number of arguments wouldn't match the passed number of arguments.

st3/lsp_utils/_client_handler/language_handler.py
jfcherng: I guess https://github.com/sublimelsp/lsp_utils/pull/38/commits/b9f81229128f1780ba2c29de4ff56175abfd9f6b resolves this at a certain degree.

st3/lsp_utils/npm_client_handler.py
jfcherng: Yes, I think it doesn't make sense to have a handler handles both notification and request. Will change.

st3/lsp_utils/_client_handler/decorator.py
```diff
@@ -0,0 +1,77 @@
...
+T_HANDLER = Callable[[Any, Any], None]
+T_SERVER_EVENTS = Union[str, Iterable[str]]
+
+_HANDLER_MARKS = {
+    "notification": "__handle_notification_events",
+    "request": "__handle_request_events",
+}
+
+
+def notification_handler(server_events: T_SERVER_EVENTS) -> Callable[[T_HANDLER], T_HANDLER]:
```
rchl, line 22: I don't understand how such a decorator can work since the argument passed to decoration is the original function, I believe, not whatever `server_events` is.

st3/lsp_utils/npm_client_handler.py
jfcherng: Resolved as of https://github.com/sublimelsp/lsp_utils/pull/38/commits/179920b79cf7cdac68ffd5ab5cc5d68203f3e7da

st3/lsp_utils/_client_handler/decorator.py
```diff
@@ -0,0 +1,77 @@
...
+
+__all__ = [
+    "notification_handler",
+    "request_handler",
+    "register_decorated_handlers",
+]
+
+# the first argument is always "self"
+T_HANDLER = Callable[[Any, Any], None]
+T_SERVER_EVENTS = Union[str, Iterable[str]]
```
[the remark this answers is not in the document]
jfcherng: I have no special opinion about this just because the codes can run with an Iterable. Limit it to be a list is okay IMO.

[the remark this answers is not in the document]
jfcherng: Thanks! I was actually out of words for naming :rofl: ... Naming is harder than cache invalidation.

st3/lsp_utils/generic_client_handler.py
jfcherng: Will investigate (my) tomorrow!

st3/lsp_utils/_client_handler/decorator.py
```diff
@@ -0,0 +1,75 @@
...
+
+            message_methods = getattr(func, handler_mark, None)  # type: Optional[List[str]]
+            if message_methods is None:
+                continue
+
+            event_registrator = getattr(api, "on_" + client_event, None)
+            if callable(event_registrator):
+                for message_method in message_methods:
+                    event_registrator(message_method, func)
+                is_registered = True
```
rchl, line 75: I think you can just `break` here instead of using `is_registered` variable for that purpose.

jfcherng: Nice catch!

st3/lsp_utils/_client_handler/decorator.py
```diff
@@ -0,0 +1,71 @@
...
+
+__all__ = [
+    "notification_handler",
+    "request_handler",
+    "register_decorated_handlers",
+]
+
+# the first argument is always "self"
+T_HANDLER = Callable[[Any, Any], None]
+T_MESSAGE_METHODS = Union[str, List[str]]
```
rchl, line 14: Types typically use CamelCase convention.
Constants (like `_HANDLER_MARKS` here) use UPPERCASE.

st3/lsp_utils/generic_client_handler.py
rchl: If you enable pushes to this PR then I'll push a fix.

jfcherng: image

Weird. I remember there used to be a "allow edit by maintainer" in the sidebar but there isn't.

rchl: Maybe it's on the bottom, where the comment box is?

jfcherng: Hmmm... can't find it

rchl:
I'd also like this to be documented in new docs.
I can handle that.

jfcherng:
MUCH appreciated for PR reviewing (many lessons learned) and documenting. And that lsp_utils refactoring :heart:

st3/lsp_utils/generic_client_handler.py
jfcherng: Hey seriously I can't find it in https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork either now or while creating a PR...

jfcherng:
The link to documents in the README is wrong btw.

st3/lsp_utils/_client_handler/decorator.py
```diff
@@ -1,6 +1,6 @@
 from ..api_wrapper_interface import ApiWrapperInterface
 from .interface import ClientHandlerInterface
-from LSP.plugin.core.typing import Any, Callable, List, Optional, Union
+from LSP.plugin.core.typing import Any, Callable, List, Optional, TypeVar, Union
```
jfcherng, line 3: There is no `TypeVar` in the master branch of LSP.
https://github.com/sublimelsp/LSP/blob/master/plugin/core/typing.py

rchl: Good point. Will add it there.

st3/lsp_utils/_client_handler/decorator.py
```diff
@@ -9,8 +9,10 @@
     "register_decorated_handlers",
 ]
 
+T = TypeVar('T')
 # the first argument is always "self"
```
jfcherng, line 13: We can annotate the first argument (`self`) as `ClientHandlerInterface`.

predragnikolic:
What do you think of a shorter name for the decorators?

`@request_handler` -> `@on_request` -> `@request`
`@notification_handler` -> `@on_notification` -> `@notification`

rchl:
> What do you think of a shorter name for the decorators?
> 
> `@request_handler` -> `@on_request`
> `@notification_handler` -> `@on_notification`

I feel that the existing naming is more in line with conventions I'm familiar with where decorator describes what it turns the function into (like `@classmethod`). `on_` prefix would kinda suggest that it listens to something on the function it decorates.

predragnikolic:
> where decorator describes what it turns the function into

When you put it on that way it makes sense to name it `request_handler` `notification_handler`.

Here is the reason on why I suggested the name change.
Because the `@notification_handler` does the same thing the same thing `api.on_notification`, I felt like that they should be named the same. 

```
def on_ready(self, api) -> None:
    api.on_notification('eslint/status', self.handle_status)

def handle_status(self, params) -> None:
    ...

# the above achieves the same result as the bellow

@on_notification("eslint/status")
def handle_status(self, params) -> None:
    ...
```

the @on_notification decorator for me would suggest that the function it decorates will be executed when the notification is received.

rchl:
I said:

> `on_` prefix would kinda suggest that it listens to something on the function it decorates.

and I think the example you made:

> api.on_notification('eslint/status', self.handle_status)

is also in line with that statement. In the case of a central `api` object, we are adding a listener to that object so I think the names being different in those cases is appropriate. It's a different way of setting up a handler.

st3/lsp_utils/_client_handler/decorator.py
rchl: That doesn't work and I don't have enough resources to figure out why is that.

Possibly due to covariance/contravariance thing.

rchl (approved):

jfcherng:
Can't wait for the next release. So many improvements for LSP-* devs :D

jfcherng:
Currently the decorator methods can't handle some special notifications like `$/progress`, `window/logMessage`. Maybe we need to refactor the implementation of this PR, to make use of https://github.com/sublimelsp/LSP/issues/2494.

rchl:
I feel like it was really only meant to handle custom requests/notifications. I think that it's consistent in that case?

It would be a bit ambiguous to specify what should happen if both LSP and a plugin handle given request at least.

jfcherng:
> It would be a bit ambiguous to specify what should happen if both LSP and a plugin handle given request at least.

If the client wants to make it into a mess, its it's issue.

PR #44, opened 2020-12-01
Respect "enabled" key in plugin configuration

rchl:


predragnikolic (approved):

rwols:
I'm not completely following this. I thought the bug was in overwriting the manual config in LSP.sublime-settings with the plugin-provided config? In the LSP package?

rchl:
With this change one can set `"enabled": false` in `LSP-json.sublime-settings` and then override to `true` in project settings for example.

I haven't tried `"enabled": false` in `LSP.sublime-settings`. It might work already or might not but why would someone need to do it there?

rwols:
> but why would someone need to do it there?

Exactly... let's forget what I said :)

PR #47, opened 2020-12-06
Add ActivityIndicator that works on the main thread

rchl:
A main-thread variant of the `ActivityIndicator` from `sublime_lib`.

Relates to https://github.com/sublimelsp/LSP-pyright/issues/36

### DIFF
--- a/st3/lsp_utils/activity_indicator.py
+++ b/st3/lsp_utils/activity_indicator.py
@@ -66,7 +66,7 @@ class ActivityIndicator:
         target: Union[StatusTarget, sublime.View, sublime.Window],
         label: Optional[str] = None,
     ) -> None:
-        self.label = label
+        self._label = label
 
         if isinstance(target, sublime.View):
             self._target = ViewTarget(target)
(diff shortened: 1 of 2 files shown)


rwols (approved):
Pretty nice!

st3/lsp_utils/activity_indicator.py
```diff
@@ -0,0 +1,135 @@
...
+    interval = 100  # type: float
+
+    _target = None  # type: Optional[StatusTarget]
+
+    def __init__(
+        self,
+        target: Union[StatusTarget, sublime.View, sublime.Window],
+        label: Optional[str] = None,
+    ) -> None:
+        self.label = label
```
jfcherng, line 69: probably we can add a explicit method to change `label` rather than `xxx.label = 'new label'`?

st3/lsp_utils/activity_indicator.py
```diff
@@ -0,0 +1,139 @@
...
+        Start displaying the indicator and animate it.
+
+        :raise ValueError: if the indicator is already running.
+        """
+        if self._state:
+            raise ValueError('Timer is already running')
+        else:
+            self._state = True
+            self._update()
+            sublime.set_timeout(self._run)
```
rwols, line 103: I'm pretty sure we don't need the extra set_timeout here

jfcherng (approved):

Issue #48, opened 2020-12-09
The ActivityIndicator may get stuck somehow


rwols:
<img width="1680" alt="image" src="https://user-images.githubusercontent.com/2431823/101691294-c9b79f80-3a6e-11eb-8460-763afc03f6a5.png">

I'm not sure exactly what caused this. But the indicator doesn't remove itself. cc @jfcherng

rwols:
They don't animate, for what it's worth.

PR #49, opened 2020-12-10
fix: activity indicator still shows after being stopped

jfcherng:
If `stop()` is called between `sublime.set_timeout(self._tick, self.interval)` and `_update()` (there is a 100ms interval between these two calls), `self._target.clear()` will be called first from `stop()` and then `self._target.set(self._render(self._ticks))` from `_update()`.

This results in the message gets "stucked" in the status bar.

---

Fixes https://github.com/sublimelsp/lsp_utils/issues/48

### DIFF
--- a/st3/lsp_utils/activity_indicator.py
+++ b/st3/lsp_utils/activity_indicator.py
@@ -118,20 +118,20 @@ class ActivityIndicator:
         :param label: The new label text
         """
         self._label = label
-        self._update()
+        if self._state:
+            self._update()
 
     def _run(self) -> None:
-        if self._state:
-            sublime.set_timeout(self._tick, self.interval)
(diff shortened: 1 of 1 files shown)


jfcherng:
The `if self._state:` check can be added in either `_update()` or `_tick()`, depending on whether we want the status message can be changed immediately when it's not running.

rchl:
> The `if self._state:` check can be added in either `_update()` or `_tick()`, depending on whether we want the status message can be changed immediately when it's not running.

I think I would put the check in `set_label` and `_tick` only as those are entry points that can be used from async thread only or from an external source. So it should be enough to check there.

rchl (approved):

rwols (approved):

PR #56, opened 2021-04-18
Fix issues with cyclic references and plugin instances not being GC'ed

### LINKED ISSUE #55: Plugin instances are not garbage-collected when using API decorators
Prompted by @rwols comment:

> Although I'm now wondering whether cyclic references are the cause of #1456

in https://github.com/sublimelsp/LSP/pull/1638#discussion_r615443545

I've checked if the plugin instances are actually destroyed and saw that those that use the "API decorator" (like https://github.com/sublimelsp/LSP-typescript/blob/60b1cca4c42cac8dca1b3a924b0824b72a37f9ab/plugin.py#L37-L37) are actually not!

rchl:
ApiWrapper receives a weak reference to the plugin instance to break the
cyclic dependency.

The handler passed to ApiWrapper methods that create API handlers
is only referenced weakly.

Fixes #55

### DIFF
--- a/st3/lsp_utils/_client_handler/abstract_plugin.py
+++ b/st3/lsp_utils/_client_handler/abstract_plugin.py
@@ -24,6 +24,7 @@ LanguagesDict = TypedDict('LanguagesDict', {
     'scopes': Optional[List[str]],
     'syntaxes': Optional[List[str]],
 }, total=False)
+ApiNotificationHandler = Callable[[Any], None]
 ApiRequestHandler = Callable[[Any, Callable[[Any], None]], None]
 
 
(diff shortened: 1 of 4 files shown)


st3/lsp_utils/_client_handler/abstract_plugin.py
```diff
@@ -22,35 +24,52 @@
...
-            session = self.__plugin.weaksession()
+            session = self.__session()
             if session:
                 session.send_response(Response(request_id, result))
 
-        def on_response(params: Any, request_id: Any) -> None:
-            handler(params, lambda result: send_response(request_id, result))
+        def on_response(handler_ref: 'ref[ApiRequestHandler]', params: Any, request_id: Any) -> None:
+            handler = handler_ref()
+            if handler:
```
rchl, line 59: Hmm, the reference is dead here. Must be missing something. Why would it go dead on me if the original `handler` method (the plugin instance method) is not dead...?

rchl: Fixed with c5fbaaeee94e8d8c99961f36039f72bdb71456bb

PR #58, opened 2021-05-11
Delete the node runtime archive if failed to extract

rchl:


predragnikolic (approved):

jfcherng:
btw there is also sha256 such as https://nodejs.org/dist/v14.16.1/SHASUMS256.txt

rchl:
> btw there is also sha256 for downloaded files such as https://nodejs.org/dist/v14.17.0/SHASUMS256.txt

Extra work. Not sure if worth it.

PR #59, opened 2021-05-18
Use Node v14 for OS greater than Windows 7

jfcherng:
I am using v14 (with env `NODE_SKIP_PLATFORM_CHECK` set to `1`) on Windows 7 without a problem (for `LSP-*` plugin usages) though.

### DIFF
--- a/st3/lsp_utils/node_runtime.py
+++ b/st3/lsp_utils/node_runtime.py
@@ -10,13 +10,21 @@ from os import path
 import os
 import shutil
 import sublime
+import sys
 import tarfile
 import urllib.request
 import zipfile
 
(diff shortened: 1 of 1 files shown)


rchl:
I really don't want to do this unless there is a good reason for it as there is currently no way to clean-up the old storage directories so the user would be left with some that take space and do nothing.

jfcherng:
Now lsp_utils installs v12 but lsp-css/html requires node >= v14.

rchl:
We should update but we should also add some code that cleans up the existing package cache directory first.

jfcherng:
I have move the discussion to https://github.com/sublimelsp/lsp_utils/issues/78

PR #63, opened 2021-06-07
Make npm server install more robust

rchl:
Before the server installation we'll create a marker file that will
only be removed on successful installation. So if the installation is
interrupted by closing ST or otherwise we'll not end up with a broken
installation that will then require re-installing the plugin.

### DIFF
--- a/st3/lsp_utils/server_npm_resource.py
+++ b/st3/lsp_utils/server_npm_resource.py
@@ -84,7 +84,7 @@ class ServerNpmResource(ServerResourceInterface):
 
     def install_or_update(self) -> None:
         try:
-            makedirs(self._server_dest, exist_ok=True)
+            makedirs(path.dirname(self._installation_marker_file), exist_ok=True)
             with open(self._installation_marker_file, 'a'):
                 pass
             if path.isdir(self._server_dest):
--- a/st3/lsp_utils/server_npm_resource.py
+++ b/st3/lsp_utils/server_npm_resource.py
@@ -85,8 +85,7 @@ class ServerNpmResource(ServerResourceInterface):
     def install_or_update(self) -> None:
         try:
             makedirs(path.dirname(self._installation_marker_file), exist_ok=True)
-            with open(self._installation_marker_file, 'a'):
-                pass
+            open(self._installation_marker_file, 'a').close()
             if path.isdir(self._server_dest):
                 shutil.rmtree(self._server_dest)
             ResourcePath(self._server_src).copytree(self._server_dest, exist_ok=True)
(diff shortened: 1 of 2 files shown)


st3/lsp_utils/server_npm_resource.py
```diff
@@ -80,13 +83,18 @@ def needs_installation(self) -> bool:
...
-        ResourcePath(self._server_src).copytree(self._server_dest, exist_ok=True)
-        dependencies_installed = path.isdir(path.join(self._server_dest, 'node_modules'))
-        if not dependencies_installed:
-            try:
+        try:
+            makedirs(self._server_dest, exist_ok=True)
+            with open(self._installation_marker_file, 'a'):
+                pass
+            if path.isdir(self._server_dest):
+                shutil.rmtree(self._server_dest)
```
predragnikolic, line 91: on line 87 the _server_dest folder is created
just to be deleted on line 91?

What is the reason behind this? :)

rchl: The `self._installation_marker_file` needs to be created at `_server_dest/..` which is also not created at this point.
So we really care about the parent of `server_dest` to be created and not the `server_dest` itself.

I admit the code is not very clear.

predragnikolic:
I tested this on windows.

1. Make sure you have LSP-pyright installed or any other LSP-* plugin that user lsp_utils and set the following setting:
```
// Settings in here override those in "lsp_utils/lsp_utils.sublime-settings"
{
	"nodejs_runtime": ["local"],
}
```
2.  Close ST.
3. remove C:\Users\Predrag\AppData\Local\Sublime Text\Package Storage\lsp_utils
4. Open ST, and open a `*.py` file and say yes to `Install Node` locally. 
5. Wait that the node installation starts.
6. Quit ST immediately after that. (so that the node installation is not finished)

7. Open ST. I get a popup to install `Install Node` locally. I confirm the modal 
and  expect the Node installation to start again,
but I see an error in the status bar `LSP-pyright: Error: Failed installing a local Node.js runtime:`

8. I close ST and open it again, there is no error now and the node installation starts again.


Is the 7 step look suspicions,
what is the expected behavior for step 7?

rchl:
What's the next log line after `LSP-pyright: Error: Failed installing a local Node.js runtime:`?

I've tried that on mac and it seems to work fine here.

predragnikolic:
> What's the next log line after LSP-pyright: Error: Failed installing a local Node.js runtime:?

It's `File is not a zip file`.

```
Package Control: Installing 1 missing package
[lsp_utils] Installing Node.js 12.20.2
LSP-pyright: Error: Failed installing a local Node.js runtime:
File is not a zip file
reloading settings Packages/User/Package Control.sublime-settings
reloading settings Packages/User/Preferences.sublime-settings
```

rchl:
Are you able to debug it a bit?
Removing those lines:
https://github.com/sublimelsp/lsp_utils/blob/0977b81198e5d7388b51ab197731347052f29403/st3/lsp_utils/node_runtime.py#L252-L253
should ensure that the "not a zip file" is not removed so that you can inspect it manually.

rchl:
I'm not able to reproduce on Windows and @predragnikolic will be away for a while so I'll just go ahead with it.

PR #65, opened 2021-06-10
Try to avoid deleting server files when in use

husanjun:
Please see   https://github.com/sublimelsp/LSP/pull/1737

### DIFF
--- a/st3/lsp_utils/generic_client_handler.py
+++ b/st3/lsp_utils/generic_client_handler.py
@@ -40,9 +40,9 @@ class GenericClientHandler(ClientHandler, metaclass=ABCMeta):
 
     @classmethod
     def cleanup(cls) -> None:
+        super().cleanup()
         if os.path.isdir(cls.package_storage()):
             shutil.rmtree(cls.package_storage())
-        super().cleanup()
 
     @classmethod
     def get_displayed_name(cls) -> str:
--- a/st3/lsp_utils/generic_client_handler.py
+++ b/st3/lsp_utils/generic_client_handler.py
@@ -6,6 +6,7 @@ from LSP.plugin import ClientConfig
 from LSP.plugin import DottedDict
 from LSP.plugin import WorkspaceFolder
 from LSP.plugin.core.typing import Any, Dict, List, Optional, Tuple
+from package_control import events
 import os
 import shutil
 import sublime
@@ -40,9 +41,14 @@ class GenericClientHandler(ClientHandler, metaclass=ABCMeta):
 
     @classmethod
     def cleanup(cls) -> None:
+
+        def run_async() -> None:
+            if os.path.isdir(cls.package_storage()):
+                shutil.rmtree(cls.package_storage())
+
+        if events.remove(cls.package_name):
+            sublime.set_timeout_async(run_async, 1000)
         super().cleanup()
-        if os.path.isdir(cls.package_storage()):
-            shutil.rmtree(cls.package_storage())
 
     @classmethod
     def get_displayed_name(cls) -> str:


rchl:
I understand why you would want that but it seems really unsafe to communicate with the server from the main thread.

What issue does it fix specifically? The issue doesn't really explain it.

rchl:
Since this will only help when stopping the server on the main thread, and we don't really want to do that, I'm thinking we could try something else.

Instead of deleting the server files, we could try to put a marker file in the server (added such functionality in a recent PR). Then only on starting the server again we would update the server files.

I'm not sure if that would still guarantee a full server stop before we touch the files but could be worth a try.

husanjun:
Yes, you only need to remove the plugin when you remove it, otherwise you just need to update it

rchl:
We can't distinguish those case though. In both situations we just get `plugin_unloaded` called.

husanjun:
https://packagecontrol.io/docs/events

I don't know if it works but I didn't try

rchl:
Interesting. Didn't remember about it.

husanjun:
@rchl My tests work fine

st3/lsp_utils/generic_client_handler.py
```diff
@@ -40,9 +41,14 @@ def setup(cls) -> None:
 
     @classmethod
     def cleanup(cls) -> None:
+
+        def run_async() -> None:
+            if os.path.isdir(cls.package_storage()):
+                shutil.rmtree(cls.package_storage())
+
+        if events.remove(cls.package_name):
+            sublime.set_timeout_async(run_async, 1000)
```
rchl, line 50: Thanks for that!

I don't trust arbitrary timeouts though. Would probably try to add some API to LSP to notify when the session has ended.

rchl: The session is not necessarily started at this point though...

husanjun: Can we use it as a temp solution？

rchl: I will look into it more soon, when I have free time. Might go with your solution if I can't find anything better?

What are your exact testing steps on Windows?

husanjun:
1. Remove the plugin through the package control tool, the console has no errors and the directory is deleted 
2. Ignore the plugin by setting, the console has no errors

rchl:
Also I'm assuming that you had server active at the time.
Also, which server have you testing with?

husanjun:
Yes,Occurs only in an active state,I use `LSP-pyright`

rwols:
This bug plagued windows users for a long time unfortunately. Maybe this diff can help in the short run. In the long run I'm hoping to fix this by running all the necessary async clean up code in the UI thread using a registered exit_handler async function: https://github.com/sublimelsp/sublime_asyncio/blob/master/globalstate.py#L147

st3/lsp_utils/generic_client_handler.py
rchl: So after checking stuff, I'm even less happy about that arbitrary timeout because there is no guarantee that the session will shutdown in 1s. It could easily take longer.

I guess this has a chance of improving situation a bit so I'm not against it but I wonder if we should investigate a more proper solution for this. Especially considering that @rwols has some plans for it for the 3.8 runtime already (which IMO is still quite far away).

rchl:
> In the long run I'm hoping to fix this by running all the necessary async clean up code in the UI thread using a registered exit_handler async function: [https://github.com/sublimelsp/sublime_asyncio/blob/master/globalstate.py#L147](https://github.com/sublimelsp/sublime_asyncio/blob/master/globalstate.py?rgh-link-date=2021-06-12T22%3A56%3A59Z#L147)

Wouldn't that potentially result in server getting called from two different threads? Maybe even deadlocks?

rwols:
No, because the _entire loop_ stops, and then runs that single async function on the main thread. Potentially, other async operations are also handled in-between the suspension points of the exit handler.

st3/lsp_utils/generic_client_handler.py
rchl: So this is the case when package is uninstalled.

When package gets updated, we will no longer delete the files on disabling the package with the idea that the server will get updated on starting the session after updating. But in that case it could also happen that the old session hasn't shut down yet when the new session is starting. This is even less likely, I guess, since updates typically take more time and so the server has more time to shut down. So it's probably OK to just merge it for now.

rchl (approved):
We'll try it out.

rchl:
I've been trying uninstalling LSP-pyright on Windows when testing https://github.com/sublimelsp/LSP-pyright/issues/75 and gotten:

```
ignored packages updated to: ["LSP-pyright", "Vintage"]
unloading python 3.3 plugin LSP-pyright.commands
reloading settings Packages/LSP-pyright/LSP-pyright.sublime-settings
unloading python 3.3 plugin LSP-pyright.plugin
  File "C:\Program Files\Sublime Text\Lib\python33\sublime_plugin.py", line 243, in unload_module
    module.plugin_unloaded()
  File "C:\Users\IEUser\AppData\Roaming\Sublime Text\Installed Packages\LSP-pyright.sublime-package\plugin.py", line 15, in plugin_unloaded
    LspPyrightPlugin.cleanup()
  File "C:\Users\IEUser\AppData\Roaming\SUBLIM~1\Packages\LSP_UT~1\st3\lsp_utils\generic_client_handler.py", line 44, in cleanup
    shutil.rmtree(cls.package_storage())
  File "./python3.3/shutil.py", line 478, in rmtree
  File "./python3.3/shutil.py", line 368, in _rmtree_unsafe
  File "./python3.3/shutil.py", line 368, in _rmtree_unsafe
  File "./python3.3/shutil.py", line 368, in _rmtree_unsafe
  File "./python3.3/shutil.py", line 368, in _rmtree_unsafe
  File "./python3.3/shutil.py", line 368, in _rmtree_unsafe
  File "./python3.3/shutil.py", line 377, in _rmtree_unsafe
  File "./python3.3/shutil.py", line 375, in _rmtree_unsafe
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\IEUser\\AppData\\Local\\Sublime Text\\Package Storage\\LSP-pyright\\14.17.1\\language-server\\node_modules\\pyright\\dist'
reloading settings Packages/User/Package Control.sublime-settings
reloading settings Packages/User/Preferences.sublime-settings
Package Control: The package LSP-pyright has been removed
reloading settings Packages/User/Package Control.sublime-settings
ignored packages updated to: ["Vintage"]
Error loading C:\Users\IEUser\AppData\Roaming\Sublime Text\Installed Packages\LSP-pyright.sublime-package: [Errno 13] Permission denied: 'C:\\Users\\IEUser\\AppData\\Roaming\\Sublime Text\\Installed Packages\\LSP-pyright.sublime-package'
reloading settings Packages/User/Package Control.sublime-settings
reloading settings Packages/User/Preferences.sublime-settings
reloading settings Packages/User/Preferences.sublime-settings
```

Which makes sense because nothing is stopping the server on uninstalling the package.
Even if it would, there is no guarantee that it will be stopped within one second.

PR #67, opened 2021-06-16
Ensure that local node runtime and python bindir are in the PATH

### LINKED ISSUE #64: Add local node runtime to the PATH when starting the server
When using a local node runtime, some servers (LSP-typescript) expect to find `node` on the `PATH` so that they can start a node sub-process.

We should add our local node to the `PATH`.

The `PATH` can be extended using `env` object in the settings but that's not ideal as that overrides the original `PATH`. Also, this should be done in a generic way for all node-based servers.

The relevant code is at https://github.com/sublimelsp/LSP/blob/main/plugin/core/types.py#L663-L664 but I don't have immediate idea how to fix that in a nice way. By nice I mean so that:
 - we do it invisibly from lsp_utils, without adding anything to LSP-* configs
 - we append the path to existing `PATH` instead of replacing whole `PATH`
 - the user is still able to set his own `env.PATH` without breaking that

rchl:
Fixes #64

### DIFF
--- a/st3/lsp_utils/_client_handler/abstract_plugin.py
+++ b/st3/lsp_utils/_client_handler/abstract_plugin.py
@@ -124,6 +124,15 @@ class ClientHandler(AbstractPlugin, ClientHandlerInterface):
             configuration.command = cls.get_command()
         return None
 
+    @classmethod
+    def on_pre_start(cls, window: sublime.Window, initiating_view: sublime.View,
+                     workspace_folders: List[WorkspaceFolder], configuration: ClientConfig) -> Optional[str]:
+        extra_paths = path.pathsep.join(cls.get_additional_paths())
+        if extra_paths:
+            original_path = configuration.env.get('PATH') or ''
+            configuration.env['PATH'] = path.pathsep.join([extra_paths, original_path])
+        return None
+
     # --- ClientHandlerInterface --------------------------------------------------------------------------------------
 
     @classmethod
--- a/st3/lsp_utils/npm_client_handler.py
+++ b/st3/lsp_utils/npm_client_handler.py
@@ -106,4 +106,3 @@ class NpmClientHandler(GenericClientHandler):
         if cls.__server:
             return cls.__server.node_bin
         return ''
-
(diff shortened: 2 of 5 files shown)


st3/lsp_utils/npm_client_handler.py
```diff
@@ -61,6 +62,10 @@ def get_additional_variables(cls) -> Dict[str, str]:
         })
         return variables
 
+    @classmethod
+    def get_additional_paths(cls) -> List[str]:
+        return [path.dirname(cls._node_bin())]
```
rchl, line 67: We shouldn't need to return anything here when using node on PATH but it should hurt to include it either.

st3/lsp_utils/_client_handler/abstract_plugin.py
```diff
@@ -124,6 +124,15 @@ def can_start(cls, window: sublime.Window, initiating_view: sublime.View,
             configuration.command = cls.get_command()
         return None
 
+    @classmethod
+    def on_pre_start(cls, window: sublime.Window, initiating_view: sublime.View,
+                     workspace_folders: List[WorkspaceFolder], configuration: ClientConfig) -> Optional[str]:
+        extra_paths = path.pathsep.join(cls.get_additional_paths())
+        if extra_paths:
+            original_path = configuration.env.get('PATH') or ''
+            configuration.env['PATH'] = path.pathsep.join([extra_paths, original_path])
```
predragnikolic, line 133: Can Node that is locally installed clash with the system installed Node.JS?

For example if a user has local Node installed in ST, and a global Node in the system,
which one will be chosen to run the `node` command?

rchl: The path is only modified for the language server subprocess.

st3/lsp_utils/npm_client_handler.py
predragnikolic: Even in the subprocess, I was curious to find out if two same named commands exist in the path, which one will win... Answer: Apparently the first one that is matched:

https://superuser.com/questions/506450/if-two-separate-path-directories-contain-a-same-named-executable-how-does-windo

st3/lsp_utils/_client_handler/abstract_plugin.py
predragnikolic: Even in the subprocess, I was curious to find out if two same named commands exist in the path, which one will win... Answer: Apparently the first one that is matched:

https://superuser.com/questions/506450/if-two-separate-path-directories-contain-a-same-named-executable-how-does-windo

PR #70, opened 2021-06-18
Add smoke test against LSP-pyright

### LINKED ISSUE #69: Bug in abstract_plugin.py broke all of my LSP plugins
**Feel free to close this**; I know this has been fixed by this release: https://github.com/sublimelsp/lsp_utils/releases/tag/1.13.3

I never installed this module explicitly; it seems like it must be a dep of the `LSP` package. A small bug in the `abstract_plugin.py` file just broke `LSP-bash`, `LSP-json`, `LSP-pyright` and `LSP-typescript` for me.

If this is required by the `LSP` package, might it be better to pull it into that package? If small changes in this package can break all of Sublime LSP, it seems like it would be safer to have the code in there.

Sublime LSP [is heavily tested](https://github.com/sublimelsp/LSP/tree/main/tests), while this repo is not

rchl:
Depends on https://github.com/SublimeText/UnitTesting/pull/192 and https://github.com/SublimeText/UnitTesting/pull/194

Fixes #69


rchl:
I might be blocked here by the fact that PC will throw an error if there is clone dependency in the `Packages`. A manual run of `Package Control: Install Local Dependency` needs to be triggered to update the loader...

.github/workflows/main.yml
```diff
@@ -0,0 +1,80 @@
...
+      - run: |
+          PATH="$HOME/.local/bin:$PATH"
+          sh github.sh bootstrap
+          rm -rf "$HOME/.config/sublime-text/Packages/lsp_utils"
+          sh github.sh install_package LSP main https://github.com/sublimelsp/LSP.git
+          sh github.sh install_package LSP-pyright master https://github.com/sublimelsp/LSP-pyright.git
+          sh github.sh install_package_control
+          rm -rf "$HOME/.config/sublime-text/Packages/lsp_utils"
+          mkdir -p "$HOME/.config/sublime-text/Packages/lsp_utils"
+          cp -r ./ "$HOME/.config/sublime-text/Packages/lsp_utils"
```
predragnikolic, line 60: What is achieved by removing the lsp_utils folder,
than re-creating it again with the same files?

rchl: Trying stuff out to workaround a blocking issue. See https://github.com/SublimeText/UnitTesting/issues/193

PR #81, opened 2021-10-05
Add an option to skip "npm install" for npm servers

rchl:
Some servers, like the new version of stylelint, bundle all dependencies
and don't need to have any dependencies installed.

This is needed because having no dependencies defined results in
no "node_modules" being created and messes up the logic that detects
if the server is installed or not.

### DIFF
--- a/st3/lsp_utils/server_npm_resource.py
+++ b/st3/lsp_utils/server_npm_resource.py
@@ -27,13 +27,15 @@ class ServerNpmResource(ServerResourceInterface):
         package_storage = options['package_storage']
         storage_path = options['storage_path']
         minimum_node_version = options['minimum_node_version']
+        skip_npm_install = options['skip_npm_install']
         node_runtime = NodeRuntime.get(package_name, storage_path, minimum_node_version)
         if not node_runtime:
             raise Exception('Failed resolving the Node Runtime. Please see Sublime Text console for more information')
-        return ServerNpmResource(package_name, server_directory, server_binary_path, package_storage, node_runtime)
+        return ServerNpmResource(
+            package_name, server_directory, server_binary_path, package_storage, node_runtime, skip_npm_install)
 
     def __init__(self, package_name: str, server_directory: str, server_binary_path: str,
-                 package_storage: str, node_runtime: NodeRuntime) -> None:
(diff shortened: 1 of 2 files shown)


predragnikolic:
This can also be used for LSP-tailwindcss.

rchl:
And https://github.com/sublimelsp/LSP-leo
@dev-sptg

st3/lsp_utils/server_npm_resource.py
```diff
@@ -71,7 +74,7 @@ def get_status(self) -> int:
 
     def needs_installation(self) -> bool:
         installed = False
-        if path.isdir(path.join(self._server_dest, 'node_modules')):
+        if self._skip_npm_install or path.isdir(path.join(self._server_dest, 'node_modules')):
             # Server already installed. Check if version has changed or last installation did not complete.
             try:
                 src_hash = md5(ResourcePath(self._server_src, 'package.json').read_bytes()).hexdigest()
```
predragnikolic, line 80: @rchl do you find it reasonable to throw an human friendly exception if the `self._server_src, 'package.json'` is missing. Something like "Make sure you have a package.json file in the server directory" (or something better that you'll come up with)

Currently, the `server_directory/package.json` file is the only requirement for the NpmHandler to work,
thus throwing an exception for that sounds reasonable for me.

rchl: But the current code has an explicit handle from `FileNotFoundError` exception and re-installs the server in that case. How would throwing an exception play into that?

rchl: We could throw an exception from `install_or_update`, if `package.json` is still missing **after the server is installed** but that would probably only help someone who is authoring the package, not the end users as it shouldn't really be possible for end users to trigger that case.

predragnikolic: >  is still missing after the server is installed but that would probably only help someone who is authoring the package

True, that was my intention, to warn the author. It happened to me when I was developing LSP-tailwindcss, some kind of the error message at that time would be useful. But than again we can leave that as a potential improvement in the future.

rchl: added code to handle that

PR #83, opened 2021-11-12
Use new unittesting actions

rchl:


rchl:
@randy3k might wanna check out the failure. Looks like a type (.py.py).

randy3k:
You could use the branch `v1` (It is a convention for GitHub actions) for stability.

randy3k:
ai, I forgot that the branch v1 is not yet lsp_utils compatible.

rchl:
no worries, this can wait

randy3k:
I have just pushed the new commits to v1.

The flow could be further simplified without the repository name and package name 😄 . I needed them because I am testing the action in UnitTesting repo.

```
      - uses: actions/checkout@v2
      - uses: SublimeText/UnitTesting/actions/setup@v1
        with:
          extra-packages: |
            sublimelsp/LSP@main
            sublimelsp/LSP-pyright
      - uses: SublimeText/UnitTesting/actions/run-tests@v1
```

rchl:
Looking good now. Nice!

randy3k:
Glad to know that. At least it is worth the effort.

.github/workflows/main.yml
```diff
@@ -28,10 +33,13 @@ jobs:
       - uses: actions/checkout@v2
       - uses: SublimeText/UnitTesting/actions/setup@v1
         with:
+          sublime_text_version: ${{ matrix.sublime_text.version }}
```
randy3k, line 37: It should be `sublime-text-version`. Github action inputs typically use `-`.

.github/workflows/main.yml
```diff
@@ -14,72 +14,33 @@ on:
...
-          sh github.sh install_package_control
-          sh github.sh copy_tested_package true  # true == overwrite existing package directory
-          sh github.sh run_tests
+          sublime-text-version: ${{ matrix.sublime_text.version }}
+          extra-packages: |
+            sublimelsp/LSP@${{ matrix.sublime_text.lsp_tag }}
+            sublimelsp/LSP-pyright
+      - uses: SublimeText/UnitTesting/actions/run-tests@v1
+        with:
+          sublime-text-version: ${{ matrix.sublime_text.version }}
```
randy3k, line 43: one last nit. `sublime-text-version` is not needed for `run-test`. It is only needed in the setup step.

rchl: thanks. also just noticed looking at the output

randy3k:
By the way, I noticed that you were testing against `sublimelsp/LSP@main` for ST4. If you want to test against the released version, you just need to put the prefix down, i.e., `sublimelsp/LSP@4070-`.

PR #85, opened 2021-11-29
Don't mutate plugin settings

### LINKED ISSUE #84: "ignore_server_trigger_chars" is deprecated
"ignore_server_trigger_chars" is deprecated, and can be removed.

```
/Users/codetribe/Library/Application Support/Sublime Text/Packages/User/LSP-volar.sublime-settings:
   16:2   warning Use disabled_capabilities instead: "disabled_capabilities": {"completionProvider": {"triggerCharacters": true}} ​unknown-source:2

```

https://github.com/sublimelsp/lsp_utils/blob/master/st3/lsp_utils/_client_handler/abstract_plugin.py#L159

rchl:
Instead of mutating the actual sublime settings, modify the client
configuration from "can_start".

Also removes the `on_client_configuration_ready` API since nothing
seems to be using it and it's not ideal to use it anyway.

Fixes https://github.com/sublimelsp/LSP/pull/1907#issuecomment-981113099
Fixes #84

### DIFF
--- a/st3/lsp_utils/generic_client_handler.py
+++ b/st3/lsp_utils/generic_client_handler.py
@@ -108,7 +108,6 @@ class GenericClientHandler(ClientHandler, metaclass=ABCMeta):
     def read_settings(cls) -> Tuple[sublime.Settings, str]:
         filename = "{}.sublime-settings".format(cls.package_name)
         loaded_settings = sublime.load_settings(filename)
-        cls.on_settings_read_internal(loaded_settings)
         changed = cls.on_settings_read(loaded_settings)
         if changed:
             sublime.save_settings(filename)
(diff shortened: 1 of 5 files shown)


st3/lsp_utils/generic_client_handler.py
```diff
@@ -112,16 +112,6 @@ def read_settings(cls) -> Tuple[sublime.Settings, str]:
         changed = cls.on_settings_read(loaded_settings)
```
predragnikolic, line 112: `cls.on_settings_read_internal(loaded_settings)` also mutates the `loaded_settings`. 
It sets the `languages` key.

```
    def on_settings_read_internal(cls, settings: sublime.Settings) -> None:
        languages = settings.get('languages', None)  # type: Optional[List[LanguagesDict]]
        if languages:
            settings.set('languages', cls._upgrade_languages_list(languages))
```

rchl: Removed that too.

I thought that the LSP would complain without it but it seems like it has all the logic needed to handle old-style `languages`. Maybe that was added after being implemented in lsp_utils...

predragnikolic (approved):
This fixes the issue.

predragnikolic:
this PR also closes https://github.com/sublimelsp/lsp_utils/issues/84

PR #86, opened 2021-12-26
Extend PATH with local Node for "npm install"

rchl:
Fixes https://github.com/sublimelsp/LSP-intelephense/issues/72

### DIFF
--- a/st3/lsp_utils/npm_client_handler.py
+++ b/st3/lsp_utils/npm_client_handler.py
@@ -76,9 +76,9 @@ class NpmClientHandler(GenericClientHandler):
     def get_additional_paths(cls) -> List[str]:
         node_bin = cls._node_bin()
         if node_bin:
-            ndoe_path = path.dirname(node_bin)
-            if ndoe_path:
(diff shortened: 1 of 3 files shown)


st3/lsp_utils/npm_client_handler.py
```diff
@@ -74,7 +74,12 @@ def get_additional_variables(cls) -> Dict[str, str]:
 
     @classmethod
     def get_additional_paths(cls) -> List[str]:
-        return [path.dirname(cls._node_bin())]
+        node_bin = cls._node_bin()
+        if node_bin:
+            ndoe_path = path.dirname(node_bin)
```
predragnikolic, line 79: Small typo

rchl:
@randy3k I wonder if you have an idea what's happening with the failures. It looks like installing dependencies randomly fails. Maybe because the process is not given enough time since the amount of dependencies that are missing varies per run.

randy3k:
I have no idea what exactly the cause was. It seems to be a GitHub connection issue. Hope that time will fix it.

PR #87, opened 2022-05-18
Update bundled Node version to latest LTS version (16)

rchl:
Time to finally update to the latest Node LTS version since we'll have a package that requires it (https://github.com/sublimelsp/LSP-Grammarly).

Unfortunately we don't have a way to clean old node installations from Cache. We could add a cleanup routine in the future but deleting stuff is always a bit risky.


rchl:
@randy3k any idea why those checks are timing out?

```
======================================================================
ERROR: setUpClass (tests.test_smoketest.PyrightSmokeTests)
----------------------------------------------------------------------
  File "/home/runner/.config/sublime-text/Packages/UnitTesting/unittesting/core/py33/suite.py", line 77, in _handleClassSetUp
    yield from deferred
  File "/home/runner/.config/sublime-text/Packages/lsp_utils/tests/setup.py", line 58, in setUpClass
    'timeout': INSTALL_TIMEOUT
TimeoutError: Condition not fulfilled within 6.00 seconds
```

Debugging those issues is always a bit cumbersome so I've figured I'll ask you before I attempt that.

randy3k:
Since some tests pass while some tests fail, may be trying a larger value of timeout?

rchl:
ST4 appears to fail pretty consistently. But yeah, I will try something later.

PR #91, opened 2022-09-15
fix npm install not running when local node_modules exists

rchl:
This fixes an issue that would only potentially affect developers working on LSP-* packages.

If someone did an `npm install` in, for example, a local `LSP-pyright/language-server/` repository then lsp_utils would not run `npm install` even after dependency is bumped in `package.json` which means that ST would use old version of the server.

I'm not sure why this check for `node_modules` already existing was there. Since `install_or_update` only runs when `needs_installation` decided that it needs to be installed, it seems to make sense to do it unconditionally. But I might be forgetting some case that it optimized...

@predragnikolic this is likely the reason you had similar issue before with server not updating.

### DIFF
--- a/st3/lsp_utils/server_npm_resource.py
+++ b/st3/lsp_utils/server_npm_resource.py
@@ -101,9 +101,7 @@ class ServerNpmResource(ServerResourceInterface):
                 shutil.rmtree(self._server_dest)
             ResourcePath(self._server_src).copytree(self._server_dest, exist_ok=True)
             if not self._skip_npm_install:
-                dependencies_installed = path.isdir(path.join(self._server_dest, 'node_modules'))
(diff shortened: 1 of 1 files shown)


predragnikolic:
> I'm not sure why this check for node_modules already existing was there.

I have a theory.
Once upon a time, there was this code:
https://github.com/sublimelsp/LSP-html/blob/73b085581f2094173f45e99bc668889488ec600e/plugin.py#L17

I guess we just migrated that piece of code to lsp_utils from the past code when we were starting with LSP-* plugins.

Now that check is not necessary, because there are other ways to determine when to update the server.

rchl:
Thanks for digging. :)
Let's try it.

PR #93, opened 2022-10-04
Implement Node version requirement check using semantic range

### LINKED ISSUE #89: [Feature Request] Restrict max supported Node.js version
This feature request comes from https://github.com/TheSecEng/LSP-copilot/issues/51. The Copilot server files are compiled in WASM files, which freezes ST if the user uses Node v18. Having `minimum_node_version()` is not enough.

I am imaging something like the following:

```py
class NpmClientHandler(GenericClientHandler):
    @classmethod
    def maximum_node_version(cls) -> Optional[Tuple[int, int, int]]:
        return None
```

Or even better, deprecates `minimum_node_version()` and introduces

```py
class NpmClientHandler(GenericClientHandler):
    @classmethod
    def support_node_version(cls, node_version: Tuple[int, int, int]) -> bool:
        """Whether the given Node version is supported by this client."""
        return True

    @classmethod
    def unsupport_node_message(cls, node_version: Tuple[int, int, int]) -> str:
        return "Node.js {} is not supported. Supported versions are v16 and v17.".format('.'.join(node_version))
```

Or 

```py
class NpmClientHandler(GenericClientHandler):
    @classmethod
    def support_node_version(cls, node_version: Tuple[int, int, int]) -> Union[Literal[True], str]:
        """Whether the given Node version is supported by this client."""
        return True  # or error message
```

rchl:
Resolves #89

@jfcherng


jfcherng (approved):
Awesome! Tested working.

.github/workflows/main.yml
```diff
@@ -51,9 +51,10 @@ jobs:
           python-version: '3.8'
       - run: sudo apt update
       - run: sudo apt install --no-install-recommends -y x11-xserver-utils
-      - run: pip3 install mypy==0.790 flake8==3.8.3 yapf==0.30.0 --user
+      - run: pip3 install mypy==0.981 flake8==5.0.4 pyright==1.1.273 --user
```
jfcherng, line 54:
```suggestion
      - run: pip3 install mypy==0.982 flake8==5.0.4 pyright==1.1.273 --user
```

PR #96, opened 2023-02-23
Allow use of electron runtime for Node-based servers

rchl:
Since electron builds have [pointer compression](https://v8.dev/blog/pointer-compression) enabled, those use quite a bit less memory than the official Node.js builds.

I haven't been able to find official Node.js builds with pointer compression enabled so using Electron itself is the next best thing and is easiest to handle since those can be installed as a NPM dependency.

Electron binary is fairly big itself (about 200MB on Mac, compared to about 100MB for Node.js) since it contains fair amount of extra stuff besides just Node.js build but it's not relevant for memory usage at runtime.

For reference, on Mac, the LSP-typescript running on a VSCode repo with a bunch of files opened uses following amounts of memory:
 - 210MB when running on the official Node.js build
 - 142MB when running on Electron's Node.js build

### DIFF
--- a/st3/lsp_utils/node_runtime.py
+++ b/st3/lsp_utils/node_runtime.py
@@ -222,6 +222,8 @@ class NodeRuntimeLocal(NodeRuntime):
 
     def node_env(self) -> Dict[str, str]:
         extra_env = super().node_env()
+        # Don't use user's local NPM config.
+        extra_env.update({'NPM_CONFIG_USERCONFIG': '/dev/null'})
         if self._use_electron:
             extra_env.update({'ELECTRON_RUN_AS_NODE': 'true'})
         return extra_env
--- a/st3/lsp_utils/node_runtime.py
+++ b/st3/lsp_utils/node_runtime.py
@@ -129,6 +129,9 @@ class NodeRuntime:
         return '{}(node: {}, npm: {}, version: {})'.format(
             self.__class__.__name__, self._node, self._npm, self._version if self._version else None)
 
+    def install_node(self) -> None:
+        raise Exception('Not supported!')
+
     def node_bin(self) -> Optional[str]:
         return self._node
 
@@ -341,24 +344,10 @@ class ElectronRuntimeLocal(NodeRuntime):
         if not path.isfile(self._install_in_progress_marker_file):
             self._resolve_paths()
 
-    def _resolve_paths(self) -> None:
-        self._node = self._resolve_binary()
-        self._npm = path.join(self._base_dir, 'yarn.js')
-
-    def _resolve_binary(self) -> Optional[str]:
-        binary_path = None
-        platform = sublime.platform()
-        if platform == 'osx':
-            binary_path = path.join(self._base_dir, 'Electron.app', 'Contents', 'MacOS', 'Electron')
-        elif platform == 'windows':
-            binary_path = path.join(self._base_dir, 'electron.exe')
-        else:
-            binary_path = path.join(self._base_dir, 'electron')
-        return binary_path if binary_path and path.isfile(binary_path) else None
+    # --- NodeRuntime overrides ----------------------------------------------------------------------------------------
 
     def node_env(self) -> Dict[str, str]:
         extra_env = super().node_env()
-        # Don't use user's local NPM config.
         extra_env.update({'ELECTRON_RUN_AS_NODE': 'true'})
         return extra_env
 
@@ -382,6 +371,23 @@ class ElectronRuntimeLocal(NodeRuntime):
             # '--verbose',
         ]
         self._run_yarn(args, cwd)
+
+    # --- private methods ----------------------------------------------------------------------------------------------
+
+    def _resolve_paths(self) -> None:
+        self._node = self._resolve_binary()
+        self._npm = path.join(self._base_dir, 'yarn.js')
+
+    def _resolve_binary(self) -> Optional[str]:
+        binary_path = None
+        platform = sublime.platform()
+        if platform == 'osx':
+            binary_path = path.join(self._base_dir, 'Electron.app', 'Contents', 'MacOS', 'Electron')
+        elif platform == 'windows':
+            binary_path = path.join(self._base_dir, 'electron.exe')
+        else:
+            binary_path = path.join(self._base_dir, 'electron')
+        return binary_path if binary_path and path.isfile(binary_path) else None
 
     def _run_yarn(self, args: List[str], cwd: str) -> None:
         if not path.isdir(cwd):
--- a/st3/lsp_utils/node_runtime.py
+++ b/st3/lsp_utils/node_runtime.py
@@ -14,9 +14,7 @@ import shutil
 import sublime
 import subprocess
 import sys
-import tarfile
 import urllib.request
-import zipfile
 
 __all__ = ['NodeRuntime', 'NodeRuntimePATH', 'NodeRuntimeLocal']
 
@@ -24,7 +22,8 @@ IS_WINDOWS_7_OR_LOWER = sys.platform == 'win32' and sys.getwindowsversion()[:2] 
 
 DEFAULT_NODE_VERSION = '16.17.1'
 ELECTRON_VERSION = '22.2.0'  # includes matching 16.17.1 version of Node.js
-NODE_DIST_URL = 'https://nodejs.org/dist/v{version}/{filename}'
+ELECTRON_DIST_URL = 'https://github.com/electron/electron/releases/download/v{version}/{filename}'
+YARN_URL = 'https://github.com/yarnpkg/yarn/releases/download/v1.22.19/yarn-1.22.19.js'
 NO_NODE_FOUND_MESSAGE = 'Could not start {package_name} due to not being able to resolve suitable Node.js \
 runtime on the PATH. Press the "Download Node.js" button to get required Node.js version \
 (note that it will be used only by LSP and will not affect your system otherwise).'
@@ -86,8 +85,8 @@ class NodeRuntime:
                     path.join(storage_path, 'lsp_utils', 'node-runtime'), use_electron, npm_config)
                 try:
                     local_runtime.check_binary_present()
-                except Exception:
-                    log_lines.append(' * Not downloaded. Asking to download...')
+                except Exception as ex:
+                    log_lines.append(' * Binaries check failed: {}'.format(ex))
                     # If first (or only) runtime is "local" then install without asking.
                     if selected_runtimes[0] != 'local':
                         if not sublime.ok_cancel_dialog(
@@ -98,7 +97,6 @@ class NodeRuntime:
                         local_runtime.install_node()
                     except Exception as ex:
                         log_lines.append(' * Failed downloading: {}'.format(ex))
-                        resolved_runtime = local_runtime
                         continue
                     try:
                         local_runtime.check_binary_present()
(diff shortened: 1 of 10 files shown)


st3/lsp_utils/node_runtime.py
```diff
@@ -11,17 +11,18 @@
...
 import urllib.request
 import zipfile
 
 __all__ = ['NodeRuntime', 'NodeRuntimePATH', 'NodeRuntimeLocal']
 
-IS_MAC_ARM = sublime.platform() == 'osx' and sublime.arch() == 'arm64'
 IS_WINDOWS_7_OR_LOWER = sys.platform == 'win32' and sys.getwindowsversion()[:2] <= (6, 1)  # type: ignore
 
-DEFAULT_NODE_VERSION = '16.15.0'
+DEFAULT_NODE_VERSION = '16.17.1'
```
jfcherng, line 24: The [latest v16 release](https://nodejs.org/dist/latest-v16.x/) is v16.19.1.

rchl: I've set this version because it matches the version that electron ships with.
When using electron the plain node.js is still used for installing dependencies so it's best to use the same one.

jfcherng: make sense.

predragnikolic:
I haven't tired this,(and I wont be able to test this this weekend)
But +1 for this,

I would rather download 100MB more ONCE, and use less RAM all the time.
than to download 100MB less ONCE, and consume more RAM all the time.

I just wasn't sure if an additional setting was necessary,
I wanted to purpose to add a new `node_runtime` array value, "local_electron", but an additional setting, is also fine by me.

predragnikolic:
Second, 
Not related to this PR.
Is having this many `node_runtime` options necessary? Would it be easier if we removed the `node_runtime` and `use_electron_for_local_runtime` setting and only have one runtime. That is the **electron node** one (or the **local node**) or whatever makes more sense. 

The reason why I am saying this, is because the `system` node_runtime... while useful, is not that great when using with `nvm` (Node Version Manager). 
I use nvm all the time and by accident I had "system"  specified in the lsp_utils setting. 
I changed the node version a couple of times with nvm and after ~one month ended up with many node folders inside LSP-* plugins in Package Storage that I had to manually delete.
```
λ  ~/.cache/sublime-text/Package Storage/LSP-typescript  tree -L 2
.
├── 12.20.2
│   └── typescript-language-server
...
├── 16.15.0
│   └── typescript-language-server
└── 16.16.0
    └── typescript-language-server

```
Most users are not aware of this and I consider this not ideal.

Either way, 
I would not change the current behavior.
I just said this to share my thoughts :)

rchl:
> Is having this many `node_runtime` options necessary? Would it be easier if we removed the `node_runtime` and `use_electron_for_local_runtime` setting and only have one runtime. That is the **electron node** one (or the **local node**) or whatever makes more sense.

I suppose the approach I've chosen here is to not force anyone into downloading extra stuff since more advanced users might not like that. So both local runtime and now electron are opt-ins.

It's a choice that was made at the beginning and I don't think we should change it now.

The situation with multiple versions could be seen as a separate issue. One that already exists.

predragnikolic:
FWIW here is the contents in Pacakge Storage.
```
  ~/.cache/sublime-text/Package Storage/lsp_utils  tree -L 2        
.
└── node-runtime
    └── 16.17.1
```

rchl:
Probably the path to the electron binary is different on linux then Mac and Windows. On mac it's in `Package Storage/lsp_utils/node-runtime/16.17.1/node/lib/node_modules/electron/dist/Electron.app/...`.

For Linux I'm assuming it's in `Package Storage/lsp_utils/node-runtime/16.17.1/node/lib/node_modules/electron/dist/electron` but I guess it might be wrong.

On error the installation is deleted so it's hard to debug I guess. You might want to just do `npm i electron` in any project and check where the binary ends up in `node_modules`.

rchl:
Seems to work just fine for me on Debian 11 ARM.

jfcherng:
Just tried and this is such a awesome memory saver :)
image

predragnikolic:
> For Linux I'm assuming it's in Package Storage/lsp_utils/node-runtime/16.17.1/node/lib/node_modules/electron/dist/electron but I guess it might be wrong.
> On error the installation is deleted so it's hard to debug I guess. You might want to just do npm i electron in any project and check where the binary ends up in node_modules.

I installed electron just to verify the path
and the path looks correct `node_modules/electron/dist/electron`

Not sure why I get errors, will try to look into it when I can. I am on Ubuntu 22.10

rchl:
Based on your logs, electron was actually installed correctly so the binary should be there. You confused me by only listing 2 levels deep.

Don't see why it would fail to resolve path. Have you tried restarting? :)

predragnikolic:
Hah, not only have I confused you, I managed to confuse myself
I will no longer try to debug things in the midnight :)

Here is the real reason why it is not working,
it fails when installing electron.
```
[lsp_utils] START output of command: "install-gelectron@22.2.0"

[lsp_utils] Command output END
--- lsp_utils Node.js resolving start ---
Resolving Node.js Runtime from lsp_utils for package LSP-pyright...
 * Failed to run npm command "install -g electron@22.2.0":
npm ERR! code ENOTEMPTY
npm ERR! syscall rename
npm ERR! path /home/predragnikolic/.npm-global/lib/node_modules/electron
npm ERR! dest /home/predragnikolic/.npm-global/lib/node_modules/.electron-OkaMSsrj
npm ERR! errno -39
npm ERR! ENOTEMPTY: directory not empty, rename '/home/predragnikolic/.npm-global/lib/node_modules/electron' -> '/home/predragnikolic/.npm-global/lib/node_modules/.electron-OkaMSsrj'

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/predragnikolic/.npm/_logs/2023-03-03T16_28_00_179Z-debug-0.log
--- lsp_utils Node.js resolving end ---
```

predragnikolic:
Ok, now I have reasons on why it fails on my Laptop. 

I changed my default npm install directory to be  `~/.npm-global`. (see https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally) 
I did this to avoid using `sudo` when installing npm pacakges globally.

This function will return `None` but I expected to return a path to electron:
```
def _resolve_electron_binary(self) -> Optional[str]:
        if not self._use_electron:
            return None
        electron_dist_dir = path.join(self._resolve_lib(), 'electron', 'dist')
        binary_path = None
        platform = sublime.platform()
        if platform == 'osx':
            binary_path = path.join(electron_dist_dir, 'Electron.app', 'Contents', 'MacOS', 'Electron')
        elif platform == 'windows':
            binary_path = path.join(electron_dist_dir, 'electron.exe')
        else:
            binary_path = path.join(electron_dist_dir, 'electron')
        return binary_path if binary_path and path.isfile(binary_path) else None
``` 
The `binary_path` variable is equal to `/home/predragnikolic/.cache/sublime-text/Package Storage/lsp_utils/node-runtime/16.17.1/node/lib/node_modules/electron/dist/electron` 
but that path doesn't exist on my laptop.

I only have 2 folders in `.../lib/node_modules/`:
```
λ  ~/.cache/sublime-text/Package Storage/lsp_utils/node-runtime/16.17.1/node/lib/node_modules  ls
corepack  npm
```

Lets take a step back and see how electron is installed:

Electron is installed with this code:
```
self.run_npm(['install', '-g', 'electron@{}'.format(ELECTRON_VERSION)], cwd=self._node_dir)
```
which is basally this command `npm -g electron@22.2.0`

and because I set a custom npm directory, it is acutally installed at this location inside `.npm-global`:
```
λ  ~/.npm-global/lib/node_modules  tree -L 1
.
├── @angular
├── browser-sync
 ...
├── electron  // Here it is
 ...
└── yarn
```

predragnikolic:
I managed to fix it for myself by using the `--prefix` flag to specify the install directory,

image

rchl:
Isn't it a bad idea to have a global configuration like that if you are switching between different versions of node? That means that all would share the same cache which I imagine is not ideal.

But I will look into it more. I think the best solution might be to ignore user settings completely instead of just overriding prefix but I'd have to investigate more.

rchl:
Can you also say npm setting have you overridden exactly?

rchl:
I'm now ignoring user's config.

But I'm thinking it might be problematic if someone NEEDS to, for example, use some proxy to download npm packages...

I suppose lsp_utils could support specifying custom NPM options so that user could set those individually.

rchl:
Also added support for specifying custom npm config overrides.

predragnikolic:
The last changes fixed the issue

> Isn't it a bad idea to have a global configuration like that if you are switching between different versions of node?

Turns out nvm doesn't support the prefix option.

I found this article which helped me better understand things https://sebhastian.com/nvm-prefix-issue/

> Most of the time, the prefix option is set to avoid permission issues blocking the npm install command.
When using NVM, the prefix is not needed because NVM installs a global package in a folder that doesn’t require superuser permissions.

rchl:
I'm struggling with deciding whether the electron should be used by default.

I don't make it the default then most people will not even know that it's an option and won't benefit from lower memory usage.

If I make it the default then it will trigger additional 200MB download which could make some people unhappy, I imagine.

There is also an option of asking using a dialog but would that be a separate dialog from the existing "Download Node.js" one or just merged into the existing one? Triggering two dialogs is not what I call very friendly UX.

jfcherng:
> There is also an option of asking using a dialog but would that be a separate dialog from the existing "Download Node.js" one or just merged into the existing one? Triggering two dialogs is not what I'd call a very friendly UX.

How about a yes/no/cancel = node/electron/canel dialog?

jfcherng:
How about use a file to flag the user has made "Node.js or Electron" choice? 

- If the flag file doesn't exist and the user uses "local", we ask "Node.js or Electron" and create the flag file. 
- If the flag file exists, just don't bother the user.

rchl:
It would all just complicate things with more buttons. We would have to make text longer to explain the difference so that the user could make an informed decision.

LDAP:
The only downside of electron are the 100mb additional download? I think we can even drop node completely. If advanced users want to use node they may want to manage it themselves anyway?

jfcherng:
Unfortunately, we still need NPM.

rchl:
All our helper packages use `package-lock.json`. I don't think yarn would respect that so we'd have to update all to use yarn then?

jfcherng:
> All our helper packages use package-lock.json. I don't think yarn would respect that so we'd have to update all to use yarn then?

Maybe we can

1. `$ yarn import` to convert `package-lock.json` into `yarn.lock` ( https://classic.yarnpkg.com/blog/2018/06/04/yarn-import-package-lock/ )
2. The previous step generates empty `node_modules/` and thus we delete it.
3. `$ yarn install`

rchl:
I've managed to make it work with manually managed electron and using downloaded yarn but I'm concerned a bit because it wouldn't give people choice to use normal Node anymore. I wonder whether it would be problematic for some. Maybe some companies are more strict about using Electron?

rchl:
Actually it's problematic to manually manage Electron using github as archives are in zip format and Python's `zipfile.ZipFile` removes executable bit from all binaries. And there are many binaries in the archive (on Mac at least) so it would be a pain to manually update those...

rchl:
Worked around executable bit issue with help of https://stackoverflow.com/questions/39296101/python-zipfile-removes-execute-permissions-from-binaries but MacOS still doesn't let it run. It appears that there are differences in how symlinks are treated. In the non working one there are no symlinks...

jfcherng:
Or on MacOS, maybe we call external program to do unzip?

rchl:
Yeah, could go that route probably.

But all that would make the changes a lot more significant and risky. I'm not convinced that I would want to release that enabled when using that approach.

Now I think I should implement that but as an option and only enable by default once it was tested manually by volunteers.

jfcherng:
> But all that would make the changes a lot more significant and risky.

Yeah, it sounds like there are too many "if" branch to implement.

rchl:
OK, now there are two separate runtimes so it's less risky. Only some minor changes were done to the existing Node runtime and its version wasn't changed to avoid extra friction.

I think we can ship this as is now.

PR #97, opened 2023-03-03
fix(pip): compare requirements.txt files instead of packages

rchl:


rchl:
@LDAP I made this change to fix https://github.com/sublimelsp/LSP-ruff/pull/10.

I'm not quite sure why we were comparing individual package versions instead of the actual content of the `requirements.txt` file when checking whether server needs to be upgraded. That seemed a little over-engineered...

LDAP (approved):

Issue #109, opened 2023-10-25
Failed deleting local Electron runtime due to long path on Windows


jfcherng:
## Steps for reproduction

On Windows,

1. Use the following settings and restart ST.

```js
{
    "nodejs_runtime": ["local"],
    "local_use_electron": true,
}
```

This forces `lsp_utils` to download Electron and use it. Everything works fine.

2. Now, use the following settings and restart ST.

```js
{
    "nodejs_runtime": ["local"],
    "local_use_electron": false,
}
```

This should ask `lsp_utils` to download Node.js. But after restarting ST, I get

```
[lsp_utils] Deleting outdated Node.js runtime directory "C:\Users\jfcherng\AppData\Local\Sublime Text\Package Storage\lsp_utils\node-runtime\18.17.1"
--- lsp_utils Node.js resolving start ---
Resolving Node.js Runtime from lsp_utils for package LSP-json...
 * Binaries check failed: "node" binary not found
 * Failed downloading: [WinError 3] File not found: 'C:\\Users\\jfcherng\\AppData\\Local\\Sublime Text\\Package Storage\\lsp_utils\\node-runtime\\18.17.1\\cache\\yarn\\v6\\npm-copilot-node-server-1.11.0-7eed36a428960e570c06cff8ccd86d70450cacdf-integrity\\node_modules\\copilot-node-server\\copilot\\dist\\compiled\\darwin\\arm64\\kerberos.node'
--- lsp_utils Node.js resolving end ---
Unable to start subprocess for LSP-json
Traceback (most recent call last):
  File "C:\Users\jfcherng\AppData\Roaming\Sublime Text\Installed Packages\LSP.sublime-package\plugin/core/windows.py", line 248, in start_async
    if plugin_class.needs_update_or_installation():
  File "C:\Users\jfcherng\AppData\Roaming\Sublime Text\Lib\python33\lsp_utils\_client_handler\abstract_plugin.py", line 101, in needs_update_or_installation
    server = cls.get_server()
  File "C:\Users\jfcherng\AppData\Roaming\Sublime Text\Lib\python33\lsp_utils\npm_client_handler.py", line 130, in get_server
    'skip_npm_install': cls.skip_npm_install,
  File "C:\Users\jfcherng\AppData\Roaming\Sublime Text\Lib\python33\lsp_utils\server_npm_resource.py", line 47, in create
    node_runtime = NodeRuntime.get(package_name, storage_path, required_node_version)
[...]

PR #110, opened 2023-10-25
fix: deleting old runtime shouldn't emit an exception

jfcherng:
Resolves https://github.com/sublimelsp/lsp_utils/issues/109

That deletion shouldn't in the try...catch block because its exception is not for ` * Failed downloading: ...` but due to locked files. This is quite confusing in the error log.

`ignore_errors=True` for `shutil.rmtree()` is being added too so that at least the whole procedure can go on (install runtime, invoke servers etc).

### DIFF
--- a/st3/lsp_utils/node_runtime.py
+++ b/st3/lsp_utils/node_runtime.py
@@ -1,4 +1,5 @@
 from .constants import SETTINGS_FILENAME
+from .helpers import rmtree_ex
 from .helpers import run_command_sync
 from .helpers import SemanticVersion
 from .helpers import version_to_string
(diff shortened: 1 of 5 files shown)


st3/lsp_utils/node_runtime.py
```diff
@@ -97,13 +97,13 @@ def _resolve_node_runtime(
                                 NO_NODE_FOUND_MESSAGE.format(package_name=package_name), 'Download Node.js'):
                             log_lines.append(' * Download skipped')
                             continue
+                    # Remove outdated runtimes.
+                    if path.isdir(runtime_dir):
+                        for directory in next(os.walk(runtime_dir))[1]:
+                            old_dir = path.join(runtime_dir, directory)
+                            print('[lsp_utils] Deleting outdated Node.js runtime directory "{}"'.format(old_dir))
+                            shutil.rmtree(old_dir, ignore_errors=True)
```
rchl, line 105: Even if it shouldn't be in the original try/except block, it should probably still be in some exception block?

In other cases we `continue` on exception so it might make sense to do here also? Maybe... saying this without recalling how things works here exactly.

Also, with or without try/except, what state are we gonna end up with when this only partially deletes the runtime dir? Would this still result in things crashing on next start without auto-healing itself?

jfcherng: After some further investigation, I found that the root cause is not due to locked file. It's because the "long path (>255 chars)" issue on Windows.

In the original error msg: ` * Failed downloading: [WinError 3] File not found: 'C:\\Users\\jfcherng\\AppData\\Local\\Sublime Text\\Package Storage\\lsp_utils\\node-runtime\\18.17.1\\cache\\yarn\\v6\\npm-copilot-node-server-1.11.0-7eed36a428960e570c06cff8ccd86d70450cacdf-integrity\\node_modules\\copilot-node-server\\copilot\\dist\\compiled\\darwin\\arm64\\kerberos.node'`

The path is 268-char long. The solution is convert the path into its UNC form as stated in code comments.

rchl: Can you also pick the 60dcba28532009334b1b0ec8ea7b5f90927bc0d3 changes from this repo? I can't push to your organization

jfcherng: Done.

rchl (approved):

PR #111, opened 2023-10-27
fix: add node directory to PATH correctly on Windows

rchl:
We've added wrong node directory path to PATH on Windows thus scripts wouldn't be able to find `node` during npm install.

Resolves https://github.com/sublimelsp/LSP-intelephense/issues/93

### DIFF
--- a/st3/lsp_utils/node_runtime.py
+++ b/st3/lsp_utils/node_runtime.py
@@ -239,7 +239,6 @@ class NodeRuntimeLocal(NodeRuntime):
         self._base_dir = path.abspath(path.join(base_dir, node_version))
         self._node_version = node_version
         self._node_dir = path.join(self._base_dir, 'node')
-        self._additional_paths = [path.join(self._node_dir, 'bin')]
         self._install_in_progress_marker_file = path.join(self._base_dir, '.installing')
         self._resolve_paths()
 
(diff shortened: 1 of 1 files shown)


predragnikolic:
CASE #1 
Test Windows when I **do not have** a global node version installed

Cannot reproduce the issue in this case.
`lsp_utils` can find the local node version, even without this PR. (as long as the user pressed "Download Node.js", in the popup that says "Could not start LSP-* due to not being able to find Node.js runtime on the PATH", 
and downloaded the new Node18 version)

CASE #2
Test Windows when I **do have** a global node version installed

I haven't tested this case.

rchl:
I've reproduced this issue in UTM so you don't really have to but if you want then make sure that:
- you are using the Node and not the Electron runtime
- the `LSP-intelephense` is not already installed in Package Storage
- you are trying with latest version of `LSP-intelephense`

predragnikolic:
Thanks, following those steps I will get an error when the server is installed for the first time.
And this PR fixes that.

----

I don't seem to reproduce this issue with other LSP node based servers on windows?
Did you?

rchl:
It appears to be specific to `postinstall` script of some of the packages so yeah, it wouldn't reproduce for every server.

PR #113, opened 2023-11-14
chore: update yarn v1.22.21 for Electron

jfcherng:
Surprised it gets updated. https://github.com/yarnpkg/yarn/releases/tag/v1.22.21


rchl:
Is there any good argument for upgrading? `lsp_utils` upgrades can be disruptive (require ST restart) so I'd prefer avoiding that if there are no specific benefits in doing that.

jfcherng:
No. We can just wait until there is a meaningful commit.

PR #118, opened 2024-04-28
feat: add support for specifying path to node in settings

taylorzane:
This pull request adds support for specifying a path to a `node` binary. I've found that `system` can be unreliable when using tools like `nodenv` (even with proper `.zprofile` setup), so this change lets a user specify any node binary to be used with LSP plugins.

I've also submitted a PR to update the `lsp_utils.sublime-settings`: [sublimelsp/LSP#2464](https://github.com/sublimelsp/LSP/pull/2464)

**Caveat:** This change assumes that an `npm` binary exists in the same directory as `node`; in my research it seems that all installations of Node.js follow this principle, so it should be a non-issue.

### DIFF
--- a/st3/lsp_utils/node_runtime.py
+++ b/st3/lsp_utils/node_runtime.py
@@ -246,6 +246,7 @@ class NodeRuntimeAbsolute(NodeRuntime):
         self._base_dir = path.abspath(path.dirname(node_binary))
         self._node = path.join(self._base_dir, 'node')
         self._npm = path.join(self._base_dir, 'npm')
+        self._additional_paths = [path.dirname(self._node)] if self._node else []
 
 class NodeRuntimePATH(NodeRuntime):
     def __init__(self) -> None:
--- a/st3/lsp_utils/node_runtime.py
+++ b/st3/lsp_utils/node_runtime.py
@@ -123,6 +123,20 @@ class NodeRuntime:
                     break
                 except Exception as ex:
                     log_lines.append(' * {}'.format(ex))
+            elif path.basename(runtime_type) == 'node':
+                log_lines.append('Resolving Node.js Runtime from absolute path for package {}...'.format(package_name))
+                path_runtime = NodeRuntimeAbsolute(runtime_type)
+                try:
+                    path_runtime.check_binary_present()
+                except Exception as ex:
+                    log_lines.append(' * Failed: {}'.format(ex))
+                    continue
+                try:
+                    path_runtime.check_satisfies_version(required_node_version)
+                    resolved_runtime = path_runtime
+                    break
+                except Exception as ex:
+                    log_lines.append(' * {}'.format(ex))
         if not resolved_runtime:
             log_lines.append('--- lsp_utils Node.js resolving end ---')
             print('\n'.join(log_lines))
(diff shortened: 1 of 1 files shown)


predragnikolic:
Hello, 
Thanks for this PR.

I use [nvm](https://github.com/nvm-sh/nvm)
and I am on a Mac.

When I have the following set in lsp_utils.sublime-settings:
```json
{
	"nodejs_runtime": ["/Users/predrag/.nvm/versions/node/v20.11.1/bin/node"],
}
```

I get this error:
```
Unable to start subprocess for LSP-json
  File "/Users/predrag/Library/Application Support/Sublime Text/Lib/python38/lsp_utils/server_npm_resource.py", line 123, in install_or_update
    self._node_runtime.run_install(cwd=self._server_dest)
  File "/Users/predrag/Library/Application Support/Sublime Text/Lib/python38/lsp_utils/node_runtime.py", line 240, in run_install
    raise Exception('Failed to run npm command "{}":\n{}'.format(' '.join(args), error))
Exception: Failed to run npm command "ci --omit=dev --verbose":
env: node: No such file or directory
```

st3/lsp_utils/node_runtime.py
```diff
@@ -226,6 +240,13 @@ def npm_command(self) -> List[str]:
         return [self._npm]
 
 
+class NodeRuntimeAbsolute(NodeRuntime):
+    def __init__(self, node_binary: str):
+        super().__init__()
+        self._base_dir = path.abspath(path.dirname(node_binary))
+        self._node = path.join(self._base_dir, 'node')
+        self._npm = path.join(self._base_dir, 'npm')
```
predragnikolic, line 248:
```suggestion
        self._npm = path.join(self._base_dir, 'npm')
        self._additional_paths = [path.dirname(self._node)]
```

this will fix the `node: No such file or directory` error

taylorzane: Went ahead and pushed that change. Let me know if you're still experiencing problems.

gerardroche:
Why not just add the path to node? Instead of:

```
"/Users/predrag/.nvm/versions/node/v20.11.1/bin/node"
```

Allow setting the path as:

```
"/Users/predrag/.nvm/versions/node/v20.11.1/bin"
```

Also, filter the path for env variables and user placeholders so that paths like `~/.nvm/versions/node/v20.11.1/bin` can be accepted too, use something like:

```
def filter_path(path):
    """Return a path with user and variables expanded."""
    path = os.path.expanduser(path)
    path = os.path.expandvars(path)

    return path
```

So now you will be able to set paths like:

```
~/.nvm/versions/node/v20.11.1/bin
$HOME/.nvm/versions/node/v20.11.1/bin
```

I'm wondering if it's possible to add the path the environment `PATH`? It looks like you can do it on a client-by-client basis: see env setting in https://lsp.sublimetext.io/client_configuration/

But you don't want to have to do that for every client, you want to be able to set it globally. That I think would be a better solution: 

- add the path to your node version to env PATH for all clients
- add the path to your node version to env PATH on a client-by-client basic (looks like you can currently already do this)

Currently, I have to make nvm available at the system level. It's not ideal.

st3/lsp_utils/node_runtime.py
```diff
@@ -226,6 +240,14 @@ def npm_command(self) -> List[str]:
...
```
rchl, line 248: Won't work on Windows.

An-dz:
I think this is a way cleaner approach:
```diff
 class NodeRuntimePATH(NodeRuntime):
     def __init__(self) -> None:
         super().__init__()
-        self._node = shutil.which('node')
-        self._npm = shutil.which('npm')
+        path = settings.get('nodejs_path') or None
+        self._node = shutil.which('node', path=path)
+        self._npm = shutil.which('npm', path=path)
```
https://github.com/sublimelsp/lsp_utils/blob/e0f4692a0e91df234b4a37730d870dab7d8cc64d/st3/lsp_utils/node_runtime.py#L229-L233

PR #120, opened 2024-06-13
fix: additional paths shouldn't be duplicated every time a new session starts

jfcherng:
`configuration.env['PATH']` is bound to a server class and we just keep prepending the same additional paths into it every time when there is a new session starts (e.g., a project is opened in a new window). Eventually, it triggers this bug: https://github.com/TerminalFi/LSP-copilot/issues/163

### DIFF
--- a/st3/lsp_utils/_client_handler/abstract_plugin.py
+++ b/st3/lsp_utils/_client_handler/abstract_plugin.py
@@ -128,19 +128,19 @@ class ClientHandler(AbstractPlugin, ClientHandlerInterface):
     @classmethod
     def on_pre_start(cls, window: sublime.Window, initiating_view: sublime.View,
                      workspace_folders: List[WorkspaceFolder], configuration: ClientConfig) -> Optional[str]:
-        original_path_raw = configuration.env.get('PATH') or ''
-        if isinstance(original_path_raw, str):
-            original_paths = original_path_raw.split(path.pathsep)
-        else:
-            original_paths = original_path_raw
-
-        # To fix https://github.com/TerminalFi/LSP-copilot/issues/163 ,
-        # We don't want to add the same path multiple times whenever a new server session is created.
-        # Note that additional paths should be prepended to the original paths.
-        wanted_paths = [path for path in cls.get_additional_paths() if path not in original_paths]
-        wanted_paths.extend(original_paths)
-        configuration.env['PATH'] = path.pathsep.join(wanted_paths)
-
+        extra_paths = cls.get_additional_paths()
+        if extra_paths:
+            original_path_raw = configuration.env.get('PATH') or ''
+            if isinstance(original_path_raw, str):
+                original_paths = original_path_raw.split(path.pathsep)
+            else:
+                original_paths = original_path_raw
+            # To fix https://github.com/TerminalFi/LSP-copilot/issues/163 ,
+            # We don't want to add the same path multiple times whenever a new server session is created.
+            # Note that additional paths should be prepended to the original paths.
+            wanted_paths = [path for path in extra_paths if path not in original_paths]
+            wanted_paths.extend(original_paths)
+            configuration.env['PATH'] = path.pathsep.join(wanted_paths)
         return None
 
     # --- ClientHandlerInterface --------------------------------------------------------------------------------------
--- a/st3/lsp_utils/_client_handler/abstract_plugin.py
+++ b/st3/lsp_utils/_client_handler/abstract_plugin.py
@@ -1,6 +1,5 @@
 from .._util import weak_method
 from ..api_wrapper_interface import ApiWrapperInterface
-from ..helpers import unique_everseen
 from ..server_resource_interface import ServerStatus
 from .api_decorator import register_decorated_handlers
 from .interface import ClientHandlerInterface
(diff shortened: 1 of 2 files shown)


st3/lsp_utils/helpers.py
```diff
@@ -93,3 +95,47 @@ def log_and_show_message(message: str, additional_logs: Optional[str] = None, sh
     print(message, '\n', additional_logs) if additional_logs else print(message)
     if show_in_status:
         sublime.active_window().status_message(message)
+
+
+def unique_everseen(iterable: Iterable[_T], key: Optional[Callable[[_T], Any]] = None) -> Generator[_T, None, None]:
```
jfcherng, line 100: This is taken from https://github.com/more-itertools/more-itertools/blob/7e46c39ea5d73b80f36ef2faf266e10108029b02/more_itertools/recipes.py#L478

jfcherng:
Unittest failures are irrelevant.

st3/lsp_utils/_client_handler/abstract_plugin.py
```diff
@@ -128,12 +130,16 @@ def can_start(cls, window: sublime.Window, initiating_view: sublime.View,
...
-            original_path = configuration.env.get('PATH') or ''
-            if isinstance(original_path, list):
-                original_path = path.pathsep.join(original_path)
-            configuration.env['PATH'] = path.pathsep.join([extra_paths, original_path])
+            original_path_raw = configuration.env.get('PATH') or ''
+            if isinstance(original_path_raw, str):
+                original_paths = original_path_raw.split(path.pathsep)
+            else:
+                original_paths = original_path_raw
+            configuration.env['PATH'] = path.pathsep.join(unique_everseen(chain(extra_paths, original_paths)))
```
rchl, line 142: We just need an `if ... in` check. No need for over-engineering with some function that will clean up unrelated paths.

And also it's even potentially wrong since it's using set which doesn't guarantee order to be the same as the input order.

jfcherng: > We just need an if ... in check. No need for over-engineering with some function that will clean up unrelated paths.

Fine. Will do.

> And also it's even potentially wrong since it's using set which doesn't guarantee order to be the same as the input order.

No. It's `set` is only for checking seen or not seen. The order is maintained.

st3/lsp_utils/_client_handler/abstract_plugin.py
```diff
@@ -128,12 +130,16 @@ def can_start(cls, window: sublime.Window, initiating_view: sublime.View,
     @classmethod
     def on_pre_start(cls, window: sublime.Window, initiating_view: sublime.View,
                      workspace_folders: List[WorkspaceFolder], configuration: ClientConfig) -> Optional[str]:
-        extra_paths = path.pathsep.join(cls.get_additional_paths())
+        # To fix https://github.com/TerminalFi/LSP-copilot/issues/163 ,
+        # We don't want to add the same path multiple times whenever a new server session is created.
```
rchl, line 134: This comment could be placed in more relevant place. Now it looks like it applies to the whole method.

jfcherng: Now it should be relevant I guess.

st3/lsp_utils/_client_handler/abstract_plugin.py
```diff
@@ -128,12 +128,19 @@ def can_start(cls, window: sublime.Window, initiating_view: sublime.View,
...
+        else:
+            original_paths = original_path_raw
+
+        # To fix https://github.com/TerminalFi/LSP-copilot/issues/163 ,
+        # We don't want to add the same path multiple times whenever a new server session is created.
+        # Note that additional paths should be prepended to the original paths.
+        wanted_paths = [path for path in cls.get_additional_paths() if path not in original_paths]
+        wanted_paths.extend(original_paths)
+        configuration.env['PATH'] = path.pathsep.join(wanted_paths)
+
```
rchl, line 143: blank lines within method bodies is not in style of LSP code base

```suggestion
            original_paths = original_path_raw
        # To fix https://github.com/TerminalFi/LSP-copilot/issues/163 ,
        # We don't want to add the same path multiple times whenever a new server session is created.
        # Note that additional paths should be prepended to the original paths.
        wanted_paths = [path for path in cls.get_additional_paths() if path not in original_paths]
        wanted_paths.extend(original_paths)
        configuration.env['PATH'] = path.pathsep.join(wanted_paths)
```

st3/lsp_utils/_client_handler/abstract_plugin.py
```diff
@@ -128,12 +128,19 @@ def can_start(cls, window: sublime.Window, initiating_view: sublime.View,
...
+        original_path_raw = configuration.env.get('PATH') or ''
+        if isinstance(original_path_raw, str):
+            original_paths = original_path_raw.split(path.pathsep)
+        else:
+            original_paths = original_path_raw
+
+        # To fix https://github.com/TerminalFi/LSP-copilot/issues/163 ,
+        # We don't want to add the same path multiple times whenever a new server session is created.
+        # Note that additional paths should be prepended to the original paths.
+        wanted_paths = [path for path in cls.get_additional_paths() if path not in original_paths]
```
rchl, line 140: All this logic is not really necessary when `cls.get_additional_paths()` is empty. So like before, I'd prefer to avoid it when not needed. Either by running this conditionally or having an early return.

rchl:
thanks

PR #123, opened 2024-08-11
fix: package control can be an optional dependency

gerardroche:
This change refactors the package control dependency to be optional by catching the import error if package control is not available and making it an no-op operation.

Previously the import error at the top prevented LSP utils from working properly when package control was not available.

### DIFF
--- a/st3/lsp_utils/generic_client_handler.py
+++ b/st3/lsp_utils/generic_client_handler.py
@@ -7,7 +7,6 @@ from LSP.plugin import ClientConfig
 from LSP.plugin import DottedDict
 from LSP.plugin import WorkspaceFolder
 from LSP.plugin.core.typing import Any, Dict, List, Optional, Tuple
-from package_control import events  # type: ignore
 import os
 import sublime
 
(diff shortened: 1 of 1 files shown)


predragnikolic:
In order for lsp_utils to be installed, PC must exist. Because PC manages dependencies.

In what scenario will the user not have PC installed?

gerardroche:
I have the dependencies manually installed.

predragnikolic:
Just to conclude, you don't want to be dependent on package control?

gerardroche:
Yes. Maybe. I thought it was going to be too difficult to install things without package control. I'm surprised how easy it is. The Sublime Text packages system out of the box is very smart.

Some language servers like phpactor worked straight away (they don't depend on `lsp_utils`), others wouldn't because the exception is raised at load time. Otherwise everything seems to be working.

As far as I can tell this is the only reference to package control.

Plugins that depend on pc typically wrap related code in a try except block, but feel free to rework the pr.

- https://github.com/SublimeLinter/SublimeLinter/blob/master/sublime_linter.py#L52
- https://github.com/NeoVintageous/NeoVintageous/blob/master/plugin.py#L174

PR #124, opened 2024-10-02
fix: update optionality of the is_allowed_to_start() method

rchl:
Those are not optional. I think the reason it was like that was to support the legacy interface that is now removed.


rwols (approved):

rwols:
I’m not completely in the loop here… is it expected that the CI fails?

rchl:
I haven't looked into it but yeah, not a real issue.

PR #125, opened 2024-10-19
fix: update node to 20.18.0

rchl:
Update to latest stable Node version.

Latest Electron version is based on Node 20.18.0 which is currently matched latest Node stable version - https://releases.electronjs.org/


st3/lsp_utils/node_runtime.py
```diff
@@ -26,10 +26,10 @@
 NODE_RUNTIME_VERSION = '18.18.1'
```
predragnikolic, line 26: Do you plan to change this value also to match the electron node version?

rchl: For sure :)
thanks

jfcherng:
One thing you may or may not care is that it will causes the Node.js won't run on some old machines, which uses glibc <= 2.17. And that's why I have https://github.com/jfcherng-sublime/ST-my-settings/blob/b4b2b100d2ff8810e51633f2ace5b9997de33edf/Packages/my_plugin_38/0_lsp_utils.py#L24-L38 because some machines I work on the company are old (e.g., CentOS 7).

rchl:
> One thing you may or may not care is that it will causes the Node.js won't run on some old machines, which uses glibc <= 2.17. And that's why I have [jfcherng-sublime/ST-my-settings@`b4b2b10`/Packages/my_plugin_38/0_lsp_utils.py#L24-L38](https://github.com/jfcherng-sublime/ST-my-settings/blob/b4b2b100d2ff8810e51633f2ace5b9997de33edf/Packages/my_plugin_38/0_lsp_utils.py#L24-L38) because some machines I work on the company are old (e.g., CentOS 7).
> 
> This PR won't affect me anyway but just a fyi.

I think I'd prefer more consistent environment that consistently doesn't work rather than dynamic like that that then causes some servers to not work.

PR #131, opened 2024-11-27
ui: respect the silence as defined in user LSP settings

eugenesvk:
LSP is [planning to add](https://github.com/sublimelsp/LSP/pull/2546) a user setting allowing for an uniterrupted user experience on various less-than-critical errors.

However, lsp_utils still asks for download in case some local node-resolution failed. This PR checks whether the user has opted into no interruptions in their LSP config and auto-cancels the download and the download prompt with a custom console log message specifying which setting is responsible

LSP will notify via statusbar that the server failed, so no need for an extra status notification in this package

addresses https://github.com/sublimelsp/LSP/issues/2545

### DIFF
--- a/st4_py38/lsp_utils/node_runtime.py
+++ b/st4_py38/lsp_utils/node_runtime.py
@@ -94,6 +94,11 @@ class NodeRuntime:
                 except Exception as ex:
                     log_lines.append(' * Binaries check failed: {}'.format(ex))
                     if selected_runtimes[0] != 'local':
+                        lsp_cfg = sublime.load_settings('LSP.sublime-settings')
+                        suppress_error_dialogs = cast(bool, lsp_cfg.get('suppress_error_dialogs') or False)
+                        if suppress_error_dialogs:
+                            log_lines.append(' * Download auto-skipped due to "suppress_error_dialogs" setting')
+                            continue
                         if not sublime.ok_cancel_dialog(
                                 NO_NODE_FOUND_MESSAGE.format(package_name=package_name), 'Download Node.js'):
                             log_lines.append(' * Download skipped')


st4_py38/lsp_utils/node_runtime.py
```diff
@@ -94,6 +94,11 @@ def _resolve_node_runtime(
...
                     log_lines.append(' * Binaries check failed: {}'.format(ex))
                     if selected_runtimes[0] != 'local':
+                        lsp_cfg = sublime.load_settings('LSP.sublime-settings')
+                        suppress_error_dialogs = cast(bool, lsp_cfg.get('suppress_error_dialogs') or False)
+                        if suppress_error_dialogs:
+                            log_lines.append(' * Download auto-skipped due to "suppress_error_dialogs" setting')
+                            continue
                         if not sublime.ok_cancel_dialog(
                                 NO_NODE_FOUND_MESSAGE.format(package_name=package_name), 'Download Node.js'):
                             log_lines.append(' * Download skipped')
```
rchl, line 104: This is not strictly an error dialog but a dialog asking user to install local node instance. Doesn't seem right to me to tie it to that LSP setting.

There is already a way to prevent this dialog from appearing by setting `"nodejs_runtime": ["system"]` setting in lsp_utils settings.

eugenesvk: Well, I did want to add non-error dialog setting for technical cases like this :), but that got rejected.

I'd still like to use the local system in general, just not be block-ui-notified about any issues, so the use case is exactly the same, and I can decide whether to just ignore these messages if it's a temp issue with system node, and I don't need a dupe, or install a local node if I don't plan to install a system one.

Although in this specific case there are no commands to manually install, right? So in addition to suppressing the dialog (which is desired) this prevents any local installation (which isn't desired and would be hard to debug)

Could we add a command to install node locally and notify that this command could be used in the message?

rchl: If you want to use the system node in general then why don't you just set the `nodejs_runtime` as I've suggested? You clearly don't want to install a "local" node instance or you would do it once and never care about this dialog anymore.

eugenesvk: In general, I want to be able to use either, not just the system one. And I don't install the local one when it's a dupe, but in cases when I didn't have a system one that's exactly what I did - installed the local version via this plugin.

PR #133, opened 2025-08-19
fix: use stable storage location for node packages

rchl:
Previously the node_modules location for npm-based plugins included the node version that was being used, for example: `../Package Storage/LSP-vue/<node_version>/node_modules/..`. With this change the node version is no longer part of the path. Changed so that the path can be known statically, regardless of which version of node is being used. I need stable path for the use case described in https://github.com/sublimelsp/LSP-vue/issues/152 where `LSP-vue` needs to provide a location of the typescript plugin for use by `LSP-typescript`.

Instead of having a node version in the path, we now create a `.node_version` file which is used for the same purpose - reinstalling dependencies if the used node version changes.

### DIFF
--- a/st4_py38/lsp_utils/server_npm_resource.py
+++ b/st4_py38/lsp_utils/server_npm_resource.py
@@ -42,7 +42,7 @@ class ServerNpmResource(ServerResourceInterface):
         required_node_version = options['required_node_version']  # type: Union[str, SemanticVersion]
         skip_npm_install = options['skip_npm_install']
         # Fallback to "minimum_node_version" if "required_node_version" is 0.0.0 (not overridden).
-        if '0.0.0' == required_node_version:
+        if required_node_version == '0.0.0':
             required_node_version = minimum_node_version
         node_runtime = NodeRuntime.get(package_name, storage_path, required_node_version)
         if not node_runtime:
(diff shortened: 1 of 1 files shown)


st4_py38/lsp_utils/server_npm_resource.py
```diff
@@ -99,6 +97,11 @@ def needs_installation(self) -> bool:
                 raise Exception('Missing required "package.json" in {}'.format(self._server_src))
             src_hash = md5(src_package_json.read_bytes()).hexdigest()
             try:
+                with open(self._node_version_marker_file) as file:
+                    node_version = str(self._node_runtime.resolve_version())
+                    stored_node_version = file.read()
+                    if node_version != stored_node_version.strip():
+                        return True
```
predragnikolic, line 104: the old code read in this order:
1. read file in Packages folder
1. read file in Package Storage folder
1. compare the hash

the new code read in this order:
1. read file in Packages folder
1. **(new)** did the node version change, if so return.
1. read file in Package Storage folder
1. compare the hash

Do you thing that moving lines 100-104, 
before the  `# Server already installed` comment would make the code a slight bit better?
1. **(new)** did the node version change, if so return.
1. read file in Packages folder
1. read file in Package Storage folder
1. compare the hash

predragnikolic (approved):
I left one minor comment, the PR is good. 
Tested it and it works.

st4_py38/lsp_utils/server_npm_resource.py
rchl: I'm not sure if moving the code before `# Server already installed` would be better but I've moved the `src_hash` line down, closer to where it's used. If moving before the comment line then we'd need to duplicate the try/except block.

predragnikolic (approved):

PR #134, opened 2025-08-21
fix: update node version to latest LTS (22.18.0)

rchl:
Figured that as we're doing #133 which will cause packages to get npm packages to get reinstalled, we might as well bump Node to the latest LTS version.

While Node 24 will be LTS soon, Electron didn't adopt it yet so we're sticking with the latest version of Node 22 for consistency.

### DIFF
--- a/st4_py38/lsp_utils/node_runtime.py
+++ b/st4_py38/lsp_utils/node_runtime.py
@@ -23,11 +23,11 @@ __all__ = ['NodeRuntime']
 
 IS_WINDOWS_7_OR_LOWER = sys.platform == 'win32' and sys.getwindowsversion()[:2] <= (6, 1)  # type: ignore
 
-NODE_RUNTIME_VERSION = '20.18.0'
+NODE_RUNTIME_VERSION = '22.18.0'
 NODE_DIST_URL = 'https://nodejs.org/dist/v{version}/{filename}'
 
-ELECTRON_RUNTIME_VERSION = '33.0.0'  # includes Node.js v20.18.0
-ELECTRON_NODE_VERSION = '20.18.0'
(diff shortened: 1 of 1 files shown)


predragnikolic:
I did a quick check of all node based LSP-* packages that I know:
```
LSP-angular
LSP-astro
LSP-bash
LSP-biome
LSP-copilot
LSP-cspell
LSP-css
LSP-dockerfile
LSP-elm
LSP-eslint
LSP-Grammarly
LSP-graphql
LSP-html
LSP-intelephense
LSP-json
LSP-leo
LSP-prisma
LSP-pyright
LSP-rome
LSP-stylelint
LSP-svelte
LSP-tailwindcss
LSP-typescript
LSP-volar
LSP-vue
LSP-yaml
LSP-basedpyright
LSP-aurelia
LSP-some-sass
```

They all work. 
---

While testing this, 
I've noticed my own regression I did last week when I merged the dependabot's PRs,
the package-lock.json `"lockfileVersion"` was updated to version `3` https://github.com/search?q=org%3Asublimelsp%20%22lockfileVersion%22%3A%203%2C&type=code

The current version of yarn we use is not able to install the packages with version `3` specified... It is the same issue as https://github.com/sublimelsp/LSP-svelte/pull/157/files . 
I will shortly create a PR for those LSP-* to fix that. 
Not sure how I didn't manage to catch those errors when I tested the dependabot PR's locally.

predragnikolic (approved):
Tested ✅ 

with both `"local_use_electron": false` and `true`, 
the lsp_utils.sublime-settings looked like this: 
```
{
	"nodejs_runtime": ["local"],
	"local_use_electron": false 
}
```

PR #148, opened 2026-01-30
feat: add UvHandler and UvVenvManager for handling python project

rchl:
- `UvRunner` - runs commands through `uv`. Uses system version if installed, otherwise downloads managed version to Package Storage
- `UvVenvManager` - creates virtual environment from provided `pyproject.toml` file
- `download_file` util - downloads given URL
- `extract_archive` util - extracts archive to specified directory. Handles zip and tar.(gz|bz2|xz).
- updated NodeRuntime to use new utilities

### DIFF
--- a/st4_py38/lsp_utils/_client_handler/abstract_plugin.py
+++ b/st4_py38/lsp_utils/_client_handler/abstract_plugin.py
@@ -72,7 +72,9 @@ class ApiWrapper(ApiWrapperInterface):
         session = self.__session()
         if session:
             session.send_request(
-                Request(method, params), lambda result: handler(result, False), lambda result: handler(result, True))
+                Request[Any, Any](method, params),
+                lambda result: handler(result, False),
+                lambda result: handler(result, True))
         else:
             handler(None, True)
 


predragnikolic (commented):
1. When uv is being downloaded, should we also display that in the status bar? Right now when the node js runtime is being downloaded we do display that in the status bar.

2. If we in the future bump the uv version, it seems that that we don't have logic to remove the outdated runtime folder (like we do for NodeJS). Which is fine for now, but we will probably need to add "Remove outdated uv" later. 

This is what we did for NodeJS.
```
                    # Remove outdated runtimes.
                    if path.isdir(runtime_dir):
                        for directory in next(os.walk(runtime_dir))[1]:
                            old_dir = path.join(runtime_dir, directory)
                            print('[lsp_utils] Deleting outdated Node.js runtime directory "{}"'.format(old_dir))
                            try:
                                rmtree_ex(old_dir)
```

rchl:
> 1. When uv is being downloaded, should we also display that in the status bar? Right now when the node js runtime is being downloaded we do display that in the status bar.
> 
>     2. If we in the future bump the uv version, it seems that that we don't have logic to remove the outdated runtime folder (like we do for NodeJS). Which is fine for now, but we will probably need to add "Remove outdated uv" later.
> 
> 
> This is what we did for NodeJS.
> 
> ```
>                     # Remove outdated runtimes.
>                     if path.isdir(runtime_dir):
>                         for directory in next(os.walk(runtime_dir))[1]:
>                             old_dir = path.join(runtime_dir, directory)
>                             print('[lsp_utils] Deleting outdated Node.js runtime directory "{}"'.format(old_dir))
>                             try:
>                                 rmtree_ex(old_dir)
> ```

1. This code is really designed to run from `install_or_update` where LSP already triggers "Installing..." status text during its execution.

2. There is code that checks the version saved in `VERSION` and triggers download and install if it's different. That said, I haven't tested it yet so not sure if it will succeed to overwrite the old version.

rchl:
> 2\. There is code that checks the version saved in `VERSION` and triggers download and install if it's different. That said, I haven't tested it yet so not sure if it will succeed to overwrite the old version.

Seems to work.

Note that given how it's implemented, `UvRunner` will only be initialized and install new version when it's actually needed for something. If LSP-esphome is already installed in Package Storage then we won't trigger UvRunner.

PR #168, opened 2026-05-04
Add helper function for initializing UvVenvManager for LspPlugin

rchl:
Add helper function for initializing `UvVenvManager` within `LspPlugin`.

It automatically adds support for a root `server_path` package setting that defaults to "auto" meaning that package-managed server instance will be used but it can be overridden to use a custom server binary.

It also extends the PATH to include venv directory if managed server instance is used.

This is a breaking change for existing users of `UvVenvManager` so this will be synchronized with changes in `LSP-ruff`, `LSP-pypls` and `LSP-esphome`.


rchl:
I guess the main thing to consider is whether there is something that can be done differently about the `server_path` setting.

Currently it's:

```
    // The path to the server binary to use for starting the language server.
    // Use `auto` for the package to manage (install/update) server automatically.
    // NOTE: Change this instead of directly updating `command` - the `${server_path}` variable is dynamically resolved
    // based on this setting.
    // WARNING: Using custom binary could result in package settings not matching what server supports.
    "server_path": "auto",
```

- Is name good?
- Perhaps instead of `auto` we should use `null` when managed?
- Improvements to description (it will be repeated across many packages)?

PR #170, opened 2026-05-10
Add NodeManager

rchl:
`NodeManager` is a new take on the existing `NodeRuntime`. The idea is that it supersedes `NodeRuntime` so that we can remove the latter once it's no longer used anywhere.

It's generally very similar but with better code organization (split into `NodeManager`, `NodeRunner` and `NodeInstaller`), Electron stuff removed (superseded by Node builds with pointer compression enabled) and modernized implementation (helper for `LspPlugin`, using `Path` and `ResourcePath` in public APIs).


lsp_utils/node_runtime.py
```diff
@@ -113,15 +114,6 @@ def _resolve_node_runtime(
...
                         continue
-                    # Remove outdated runtimes.
-                    if runtime_dir.is_dir():
-                        for directory in next(os.walk(runtime_dir))[1]:
-                            old_dir = runtime_dir / directory
-                            logger.info(f'Deleting outdated Node.js runtime directory "{old_dir}"')
-                            try:
-                                rmtree_ex(old_dir)
-                            except Exception as ex:
-                                log_lines.append(f' * Failed deleting: {ex}')
```
rchl, line 124: No longer cleaning up runtimes on installing one because:
a) that wouldn't work with both `NodeRuntime` (old) and `NodeManager` (new) being used at the same time
b) it was also inconvenient when ST would pick up different versions randomly (because it failed to read PATH on start on mac for example).

predragnikolic (dismissed):

rchl:
@predragnikolic feel free to provide feedback if you have any.

Merged it because it's mostly new code that is not yet used so it's pretty safe.

PR #172, opened 2026-05-15
fix(uv): create installation marker

rchl:
Create installation marker while uv venv is being created. Can help to avoid ending up with incomplete env.


predragnikolic:
Looks like this PR tries to address https://github.com/sublimelsp/LSP-ruff/issues/154#issuecomment-4461516830

rchl:
Nah, I don't think this would help there. I think that case is more about locked files.

PR #193, opened 2026-06-04
Fix 'must await a coroutine' error

rwols:
Depends on https://github.com/sublimelsp/LSP/pull/2880

### DIFF
--- a/lsp_utils/_client_handler/abstract_plugin.py
+++ b/lsp_utils/_client_handler/abstract_plugin.py
@@ -54,16 +54,14 @@ class ApiWrapper(ApiWrapperInterface):
     @override
     def on_request(self, method: str, handler: ApiRequestHandler) -> None:
         def send_response(request_id: Any, result: Any) -> None:
-            session = self.__session()
-            if session:
-                session.send_response(Response(request_id, result))
+            if session := session self.__session():
+                session.create_task(session.send_response(Response(request_id, result)))
 
         def on_response(weak_handler: WeakMethod[ApiRequestHandler], params: Any, request_id: Any) -> None:
             if handler := weak_handler():
                 handler(params, lambda result: send_response(request_id, result))
 
-        plugin = self.__plugin()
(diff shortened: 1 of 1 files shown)


lsp_utils/_client_handler/abstract_plugin.py
```diff
@@ -54,16 +54,14 @@ def handle_notification(weak_handler: WeakMethod[ApiNotificationHandler], params
     @override
     def on_request(self, method: str, handler: ApiRequestHandler) -> None:
         def send_response(request_id: Any, result: Any) -> None:
-            session = self.__session()
-            if session:
-                session.send_response(Response(request_id, result))
+            if session := session self.__session():
+                session.create_task(session.send_response(Response(request_id, result)))
```
rchl, line 58: I see that this triggers following error in the workflow.

> /Users/runner/work/LSP/LSP/lsp_maintainers/repositories/lsp_utils/lsp_utils/_client_handler/abstract_plugin.py:59:17 - error: Result of async function call is not used; use "await" or assign result to variable (reportUnusedCoroutine)
> Error: Result of async function call is not used; use "await" or assign result to variable (reportUnusedCoroutine)

We can adapt the rules where it makes sense.
Or use a type ignore comment instead.
Or assign to a dummy variable `_ = session.create_task(...)`

rchl: Though this code has syntax error also
```suggestion
            if session := self.__session():
                session.create_task(session.send_response(Response(request_id, result)))
```

rchl:
I've actually just removed those in #194 since nothing is using those and the idea is to remove as much as possible of this code.

lsp_utils/_client_handler/abstract_plugin.py
rwols: Right… I modified this file with basedpyright and ruff turned off while working on the tests in LSP

